PR #4, opened 2025-06-12
feat(plugin): initial stub

joe0BAB:
This builds on #23 #22 #21 #20 #18 and wires everything together into the stub code that exposes the `Plugin` interface and is to be used for plugin authors. It contains all the logic for plugins to set themselves up correctly when discovered by the plugin engine or started standalone (and to then connect to an existing running plugin engine).

It also wires in the secret engine resolver interface.


pkg/stub/stub.go
```diff
@@ -33,15 +37,37 @@ type Config struct {
 
 // Stub is the interface the stub provides for the plugin implementation.
 type Stub interface {
+	// Run starts the plugin then waits for the plugin service to exit, either due to a
+	// critical error or an explicit call to Stop(). Once Run() returns, the plugin can be
+	// restarted by calling Run() or Start() again.
```
Benehiko: So doing another `Run()` call would restart it? I don't see where that is checked. I only see it error under `runnable` https://github.com/docker/secrets-engine/blob/d0b590c44d19e1b391734a4cc38d4e2201316e1b/pkg/stub/runner.go#L78-L80

Benehiko (commented):
Minor nit, I think the majority of this looks good

pkg/stub/stub.go
joe0BAB: I'll add a test

pkg/stub/stub.go
```diff
@@ -33,15 +39,30 @@ type Config struct {
 
 // Stub is the interface the stub provides for the plugin implementation.
 type Stub interface {
+	// Run starts the plugin then waits for the plugin service to exit, either due to a
+	// critical error or by cancelling the context. Once Run() returns, the plugin can be
+	// restarted by calling Run() again.
```
Benehiko: Still a bit confused by the language here. 

If I wait for `Run()` to return, then it means the plugin has been stopped. So I cannot "restart" something that has been stopped, I can only "start" it, right?

So perhaps we should just update the godoc to reflect that.

```suggestion
	// Run starts the plugin then waits for the plugin service to exit, either due to a
	// critical error or by cancelling the context. Calling Run() while the plugin is running,
	// will result in an error. After the plugin service exits, Run() can safely be called again. 
```

joe0BAB: Good call out! Will change it.

Benehiko (approved):

PR #7, opened 2025-06-13
feat(api): add resolver API

joe0BAB:
Added the equivalent of the resolver interface to the protobuf API.


pkg/api/resolver/v1/api.proto
```diff
@@ -40,8 +39,30 @@ message ConfigureRequest {
...
-message ShutdownResponse {}
\ No newline at end of file
+message ShutdownResponse {}
+
+service ResolverService {
+  // Resolve a secret by its ID.
+  rpc GetSecret(GetSecretRequest) returns (GetSecretResponse);
+}
+
+message GetSecretRequest {
```
joe0BAB, line 58: I'm only adding the minimal to get unblocked on setting up the plugin system. I'll add all fields later (as we'll probably need to adjust anyway still).

Benehiko (approved):

PR #8, opened 2025-06-13
feat(api): add context field

joe0BAB:
The `GetSecret` API method was missing a `context.Context` argument that allows to cancel requests.


pkg/secrets/secrets.go
```diff
@@ -32,5 +33,5 @@ type Envelope struct {
 }
 
 type Resolver interface {
-	GetSecret(request Request) (Envelope, error)
+	GetSecret(ctx context.Context, request Request) (Envelope, error)
```
joe0BAB, line 36: This is the main change. Everything else is just propagating things through.

Benehiko (approved):

PR #12, opened 2025-06-16
feat(plugin): implement plugin side configure service

joe0BAB:
This serves runs inside the plugin. It should be called once by the engine to pass over configuration to the plugin.


pkg/stub/configure_test.go
```diff
@@ -0,0 +1,144 @@
...
+	dynamicCfg := cfgFromEngine{
+		engineName:     expCfg.engine,
+		engineVersion:  expCfg.version,
+		requestTimeout: 500 * time.Millisecond,
+	}
+	mockCfg := &Config{Version: "v1.2.3", Pattern: "test-*"}
+	mock := newMockConfigure(t, mockCfg, nil, expCfg)
+	service := newConfigureService("pluginX", mock, nil)
+	cfg, err := service.GetConfig()
+	assert.NoError(t, err)
```
joe0BAB: For now I've defined calling this before the configuration has gone through does not return an error but a nil config. This is a bit smelly though. Maybe a dedicated error would be better that we could check on the caller side, eg with `errors.Is` or so?

joe0BAB: Never mind, I'll change it directly to return an error and make it more idiomatic Go.

pkg/stub/configure.go
```diff
@@ -0,0 +1,121 @@
...
+	engineName     string
+	engineVersion  string
+	requestTimeout time.Duration
+}
+
+type cfgService struct {
+	m                    sync.Mutex
+	pluginName           string
+	configure            ConfigureInterface
+	shutdown             func(context.Context)
```
joe0BAB, line 45: This is not part of the `ConfigureInterface` as the `onShutdown` callback is optional / could be nil.

pkg/stub/configure.go
```diff
@@ -0,0 +1,121 @@
...
+
+	resolverv1 "github.com/docker/secrets-engine/pkg/api/resolver/v1"
+	"github.com/docker/secrets-engine/pkg/api/resolver/v1/resolverv1connect"
+)
+
+var (
+	errPluginNotConfiguredYet = errors.New("plugin not configured yet")
+)
+
+var _ = (resolverv1connect.PluginServiceHandler)((*cfgService)(nil))
```
Benehiko, line 21:
```suggestion
var _ resolverv1connect.PluginServiceHandler = &cfgService{}
```

The compiler is smart enough to discard this.

pkg/stub/configure.go
Benehiko: i wonder if that should be a requirement? perhaps we should enforce lifecycle management from the secrets engine on plugins. We could always do an opt-out option.

Another option is to declare an interface like `LifecycleManager` and have the plugin optionally implement it.

pkg/stub/configure.go
```diff
@@ -0,0 +1,121 @@
...
+
+// ConfigureInterface handles Configure API request.
+type ConfigureInterface interface {
```
Benehiko, line 24:
```suggestion
type Configurer interface {
```

I don't think interfaces should contain the `Interface` part in the name.

https://go.dev/doc/effective_go#interface-names

pkg/stub/configure.go
```diff
@@ -0,0 +1,121 @@
...
+	config               *cfgFromEngine
+	pluginCfgCallbackErr error
+}
+
+func newConfigureService(pluginName string, configure ConfigureInterface, shutdown func(context.Context)) *cfgService {
+	return &cfgService{
+		pluginName: pluginName,
+		configure:  configure,
+		shutdown:   shutdown,
+		done:       make(chan struct{}),
```
Benehiko, line 56: I'm honestly not so sure about this channel. For the most part we can replace it with a `sync.AtomicBool`.

I'm also not sure what the purpose of `WaitUntilConfigured` is, since `Configure` is a blocking function that returns the response.

pkg/stub/configure.go
```diff
@@ -0,0 +1,121 @@
...
+		Pattern: proto.String(cfg.Pattern),
+	}.Build()), nil
+}
+
+func (s *cfgService) Shutdown(ctx context.Context, _ *connect.Request[resolverv1.ShutdownRequest]) (*connect.Response[resolverv1.ShutdownResponse], error) {
+	if s.shutdown != nil {
+		s.shutdown(ctx)
+	}
+	return connect.NewResponse(&resolverv1.ShutdownResponse{}), nil
+}
```
Benehiko, line 98: How can we shut down if we have not configured anything yet? Should we check?

pkg/stub/configure.go
```diff
@@ -0,0 +1,121 @@
...
+	select {
+	case <-ctx.Done():
+		if err := ctx.Err(); err != nil && !errors.Is(err, context.Canceled) {
+			return err
+		}
+		return nil
+	case <-s.done:
+	}
+	return s.pluginCfgCallbackErr
+}
```
Benehiko, line 110: This method seems to be adding a lot of complexity, I think we can remove it entirely.
```suggestion
```

pkg/stub/configure.go
```diff
@@ -0,0 +1,121 @@
...
+func newConfigureService(pluginName string, configure ConfigureInterface, shutdown func(context.Context)) *cfgService {
+	return &cfgService{
+		pluginName: pluginName,
+		configure:  configure,
+		shutdown:   shutdown,
+		done:       make(chan struct{}),
+	}
+}
+
+func (s *cfgService) Configure(ctx context.Context, c *connect.Request[resolverv1.ConfigureRequest]) (*connect.Response[resolverv1.ConfigureResponse], error) {
```
Benehiko, line 60: If you want to make `Configure` a non-blocking function, you can change the `Configure` signature to return a channel instead that would get the config pushed to it. But I'm not sure why we want to go so complicated with `Configure`.

To me the logic seems to be:
1. create `cfgService`
2. set `cfgService.config` before calling `Configure`
3. call `Configure` from the `Configure Interface`
4. block with a response after connecting to the plugin

In step 2 we already set `config` without needing to call `Configure`. That forces us to use the channel (or atomic.Bool), since we can fetch the config inside a separate goroutine.

I don't see why we can't move `Shutdown()` to an interface and have a normal function here that accepts an instance of `Configure Interface` and calls `Configure` and does the connection. Then have another function with `Shutdown` that accepts a `LifecycleManager interface` that calls the shutdown and returns the connect response.

joe0BAB:
no longer needed, solved by re-architecturing

PR #16, opened 2025-06-17
store/keychain: cross-platform test cli and more tests

Benehiko:
This patch adds an easy to use CLI to the `keychain` package. It further cleans up the keychain tests.


store/keychain/cmd/main.go
```diff
@@ -0,0 +1,123 @@
...
+			return kc.Delete(cmd.Context(), id)
+		},
+	}
+	root := &cobra.Command{}
+	root.AddCommand(list, save, get, erase)
+
+	return root, nil
+}
+
+func main() {
```
joe0BAB, line 116: non-blocking(maybe for follow up PR): Maybe we should document somewhere what this binary is about, what it's for and how to build/run it?

store/keychain/cmd/main.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/docker/secrets-engine/store"
+	"github.com/docker/secrets-engine/store/keychain"
+	"github.com/docker/secrets-engine/store/mocks"
+	"github.com/spf13/cobra"
+)
+
+func newCommand() (*cobra.Command, error) {
+	kc, err := keychain.New(
+		"io.docker.Secrets",
+		"docker-example-cli",
```
joe0BAB, line 21: not sure this matters much, but I'd be inclined to be more specific here, e.g., `docker-local-secrets-example-cli` or something as `docker-example-cli` without context feels very generic (the circumstances of the code would give more context, but with a more explicit name it wouldn't even need that)

store/keychain/cmd/main.go
```diff
@@ -0,0 +1,123 @@
...
+			if err != nil {
+				return err
+			}
+
+			for k, v := range secrets {
+				vv, err := v.Marshal()
+				if err != nil {
+					return err
+				}
+				fmt.Printf("\nID: %s\nValues: %s\n", k, vv)
```
joe0BAB: Is it really `Values` and not `Value`?

store/keychain/cmd/main.go
```diff
@@ -0,0 +1,123 @@
...
+				return nil
+			}
+			if err != nil {
+				return err
+			}
+
+			for k, v := range secrets {
+				vv, err := v.Marshal()
+				if err != nil {
+					return err
```
joe0BAB: non-blocking: I don't know really how you intend to use that CLI here, but it might be more useful to just collect the error here, print something and continue and return all errors later vs aborting early

store/keychain/cmd/main.go
```diff
@@ -0,0 +1,123 @@
...
+		},
+	}
+
+	var (
+		username string
+		password string
+	)
+	save := &cobra.Command{
+		Use:     "save",
+		Aliases: []string{"set"},
```
joe0BAB: `store`?

store/keychain/cmd/main.go
```diff
@@ -0,0 +1,123 @@
...
+			}
+			secret, err := kc.Get(cmd.Context(), id)
+			if err != nil {
+				return err
+			}
+			val, err := secret.Marshal()
+			if err != nil {
+				return err
+			}
+			fmt.Printf("Secret:\nID:%s\nValues:%s\n", id.String(), val)
```
joe0BAB: same as above, `Values` vs `Value`

store/keychain/keychain_test.go
```diff
@@ -1,50 +1,70 @@
...
 			Username: "bob",
 			Password: "bob-password",
 		}
+		t.Cleanup(func() {
+			require.NoError(t, ks.Delete(t.Context(), id))
+		})
 		require.NoError(t, ks.Save(t.Context(), id, creds))
 	})
 
 	t.Run("get credential", func(t *testing.T) {
```
joe0BAB, line 64:
```suggestion
	t.Run("save/get credential", func(t *testing.T) {
```
aren't we testing both?

store/keychain/keychain_test.go
```diff
@@ -67,22 +92,28 @@ func TestKeychain(t *testing.T) {
...
 		_, err := ks.Get(t.Context(), id)
 		require.ErrorIs(t, err, store.ErrCredentialNotFound)
 	})
+
+	t.Run("delete non-existent credential", func(t *testing.T) {
+		ks := setupKeychain(t)
+		id := store.ID("com.test.test/test/does-not-exist")
+		require.NoError(t, id.Valid())
+		require.NoError(t, ks.Delete(t.Context(), id))
+	})
```
joe0BAB, line 144: I think it could be good to add some tests on the failure cases, eg
- invalid IDs 
- marshaling/unmarshaling errors

(and then test against all methods, save/get/delete/list)

Benehiko: yeah I agree, I'll add some more tests then

store/keychain/keychain_test.go
Benehiko: we kind of are, i guess having a separate run at least for `get` might have better debugging / failure logs?

store/keychain/cmd/main.go
Benehiko: it's kind of supposed to be generic - it's just used to basically test the keychain library manually.

store/keychain/README.md
```diff
@@ -30,3 +30,30 @@ func main() {
...
+## Example CLI
+
+The `keychain` package also contains an example CLI tool to test out how a real
+application might interact with the host keychain.
+
+You can build the CLI by running `go build` inside the `store/` root directory.
+
+```console
+⋊> ~/G/s/store on keychain ⨯ go build -o keychain-cli ./keychain/cmd/
+⋊> ~/G/s/store on keychain ⨯ ./keychain-cli                                                                                                                       13:24:48
```
joe0BAB:
```suggestion
$ go build -o keychain-cli ./keychain/cmd/
$ ./keychain-cli
```

store/keychain/README.md
```diff
@@ -30,3 +30,30 @@ func main() {
...
+  delete
+  get
+  help        Help about any command
+  list
+  store
+
+Flags:
+  -h, --help   help for this command
+
+Use " [command] --help" for more information about a command.
```
joe0BAB: I'd remove that, it gets outdated easily and doesn't bring that much of additional value having it here

store/keychain/cmd/main.go
joe0BAB: But wouldn't it appear in the keychain, eg, if I'd open the mac keychain app, wouldn't I see it there? I was mostly wondering from that perspective, if it would be preferable to have something less generic than `docker-example-cli` show there.

store/keychain/cmd/main.go
```diff
@@ -0,0 +1,126 @@
...
+			val, err := secret.Marshal()
+			if err != nil {
+				return err
+			}
+			fmt.Printf("Secret:\nID:%s\nValue:%s\n", id.String(), val)
+			return nil
+		},
+	}
+
+	erase := &cobra.Command{
```
joe0BAB, line 98:
```suggestion
	delete := &cobra.Command{
```
nit

store/keychain/keychain_test.go
joe0BAB: I don't have a strong opinion here, up to you

store/keychain/keychain_test.go
```diff
@@ -67,22 +94,113 @@ func TestKeychain(t *testing.T) {
...
+		})
+		require.NoError(t, kc.Save(t.Context(), id, &mustUnmarshalError{}))
+		_, err = kc.GetAll(t.Context())
+		require.ErrorContains(t, err, "i am failing on purpose")
+	})
+}
+
+type mustMarshalError struct{}
+
+// Marshal implements store.Secret.
```
joe0BAB: nit: remove, line 192 already says that

store/keychain/keychain_test.go
```diff
@@ -67,22 +94,113 @@ func TestKeychain(t *testing.T) {
...
+func (m *mustMarshalError) Marshal() ([]byte, error) {
+	return nil, errors.New("i am failing on purpose")
+}
+
+// Unmarshal implements store.Secret.
```
joe0BAB: nit: remove, line 192 already says that

store/keychain/keychain_test.go
```diff
@@ -67,22 +94,113 @@ func TestKeychain(t *testing.T) {
...
+func (m *mustMarshalError) Unmarshal(data []byte) error {
+	return nil
 }
+
+var _ store.Secret = &mustMarshalError{}
```
joe0BAB: moving that up might save us the comments :D

store/keychain/keychain_test.go
```diff
@@ -67,22 +94,113 @@ func TestKeychain(t *testing.T) {
...
+func (m *mustUnmarshalError) Marshal() ([]byte, error) {
+	return []byte("eeyyy"), nil
+}
+
+// Unmarshal implements store.Secret.
+func (m *mustUnmarshalError) Unmarshal(data []byte) error {
+	return errors.New("i am failing on purpose")
+}
+
+var _ store.Secret = &mustUnmarshalError{}
```
joe0BAB: same as above

joe0BAB (approved):
LGTM (just a few non-blocking nits)

store/keychain/cmd/main.go
Benehiko: i mean, this is is never automatically executed - so that shouldn't be a problem.

store/keychain/cmd/main.go
Benehiko: delete is a keyword in Go, that's why i kept it :see_no_evil:

PR #17, opened 2025-06-17
feat(sdk): helper functions to parse plugin names

joe0BAB:
Plugins that are discovered from the filesystem should be named in the format `<index>-<name>`. This PR provides helper functions for parsing.


pkg/api/util.go
```diff
@@ -0,0 +1,48 @@
...
+	split := strings.SplitN(name, "-", 2)
+	if len(split) < 2 {
+		return "", "", fmt.Errorf("invalid plugin name %q, idx-pluginname expected", name)
+	}
+
+	if err := CheckPluginIndex(split[0]); err != nil {
+		return "", "", err
+	}
+
+	return split[0], split[1], nil
```
joe0BAB, line 19: We don't convert to number here as we never need this to be an actual number in our code (even for sorting/ordering the plugins string is fine). Plus keeping it as string makes using it for logging more trivial.

pkg/api/util.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+func ParsePluginName(name string) (string, string, error) {
+	split := strings.SplitN(name, "-", 2)
+	if len(split) < 2 {
+		return "", "", fmt.Errorf("invalid plugin name %q, idx-pluginname expected", name)
```
Benehiko:
```suggestion
		return "", "", fmt.Errorf("invalid plugin name %q, <[0-9][0-9]>-<pluginname> expected", name)
```

Benehiko: maybe we could tell the user exactly how the format should look like?

joe0BAB: done

Benehiko (approved):

PR #18, opened 2025-06-17
feat(sdk): static stub configuration

joe0BAB:
There's a set of static configuration on the stub that either
- gets set from the secret engine for automatically launched plugins
- gets set via options for manually launched plugins
- gets implicitly pulled from the binary name for manually launched plugins

As this is somewhat implicit but very externally facing, I want to make sure everything is well tested and documented.


pkg/adaptation/adaptation.go
```diff
@@ -0,0 +1,17 @@
+package adaptation
+
+import "time"
+
+const (
+	// DefaultSocketPath is the default socket path for external plugins.
+	DefaultSocketPath = "/var/run/secrets-engine/engine.sock"
```
Benehiko: hmm, I wonder if it would make sense to run the socket as root? 

I guess this is up for debate and maybe we should prefer a "by default rootless" approach. 
In future the secrets engine integration with containerd would probably run at whatever level containerd is running, but its plugins would never need root access.

pkg/stub/stub.go
```diff
@@ -2,6 +2,9 @@ package stub
 
 import (
 	"context"
+	"time"
+
+	"github.com/sirupsen/logrus"
```
Benehiko, line 7: I wonder if we can replace logrus in favor of a better logger - this is a discussion for another time, but just thought I would call it out.

pkg/stub/config.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil, 0, fmt.Errorf("invalid registration timeout %q: %w", timeoutStr, err)
+	}
+	return &identity{name: name, idx: idx}, timeout, nil
+}
+
+func isPluginEnvSet() bool {
+	name := os.Getenv(adaptation.PluginNameEnvVar)
+	idx := os.Getenv(adaptation.PluginIdxEnvVar)
+	timeoutStr := os.Getenv(adaptation.PluginRegistrationTimeoutEnvVar)
+	return name != "" || idx != "" || timeoutStr != ""
```
Benehiko: shouldn't these be `&&`?

```suggestion
	return name != "" && idx != "" && timeoutStr != ""
```

pkg/stub/config.go
```diff
@@ -0,0 +1,170 @@
...
+	socketPath          string
+	conn                net.Conn
+	registrationTimeout time.Duration
+}
+
+func newCfg(p Plugin, opts ...Option) (*cfg, error) {
+	var identity identity
+	timeout := adaptation.DefaultPluginRegistrationTimeout
+	if isPluginEnvSet() {
+		i, t, err := getCfgFromEnv()
```
Benehiko:
```suggestion
                var err error
		identity, timeout, err = getCfgFromEnv()
```

pkg/stub/config.go
```diff
@@ -0,0 +1,170 @@
...
+		if err != nil {
+			return nil, err
+		}
+		identity = *i
+		timeout = t
```
Benehiko:
```suggestion

```

pkg/stub/config.go
```diff
@@ -0,0 +1,170 @@
...
+type cfg struct {
+	plugin Plugin
+	identity
+	socketPath          string
+	conn                net.Conn
+	registrationTimeout time.Duration
+}
+
+func newCfg(p Plugin, opts ...Option) (*cfg, error) {
+	var identity identity
```
Benehiko:
```suggestion
	var identity *identity
```

pkg/stub/config.go
```diff
@@ -0,0 +1,170 @@
...
+		i.name = filepath.Base(os.Args[0])
+		return &i, nil
+	}
+	idx, name, err := api.ParsePluginName(filepath.Base(os.Args[0]))
+	if err != nil {
+		return &i, err
+	}
+	i.idx = idx
+	i.name = name
+	return &i, nil
```
Benehiko: the `complementIdentity` function doesn't make much sense to me.

Especially since we are doing `os.Args[0]`. Feels like it's doing too much and in a non-obvious way. I'd also prefer that we stay away from things like `os.Args[0]`. Making this function testable requires magic, which makes the test code less readable in my opinion.

Could we perhaps refactor this and set explicitly what the fallback parameters are or perhaps split it up into digestible pieces.

pkg/stub/config.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if idx == "" {
+		return nil, 0, errPluginIdxNotSet
+	}
+	if err := api.CheckPluginIndex(idx); err != nil {
+		return nil, 0, fmt.Errorf("invalid plugin index %q: %w", idx, err)
+	}
+	if timeoutStr == "" {
+		return nil, 0, errPluginRegistrationTimeoutNotSet
+	}
```
Benehiko: you can inline the checks

```suggestion
       var (
           name string
           idx string
           timeoutStr string
        )
        
        if name = os.Getenv(adaptation.PluginNameEnvVar); name == "" {
            return nil, 0, errPluginNameNotSet
        }
        
        if idx = os.Getenv(adaptation.PluginIdxEnvVar); idx == "" {
            return nil, 0, errPluginIdxNotSet
        }
        
        if timeoutStr = os.Getenv(adaptation.PluginRegistrationTimeoutEnvVar); timeoutStr == "" {
		return nil, 0, errPluginRegistrationTimeoutNotSet
	}

	if err := api.CheckPluginIndex(idx); err != nil {
		return nil, 0, fmt.Errorf("invalid plugin index %q: %w", idx, err)
	}
```

pkg/stub/config.go
```diff
@@ -0,0 +1,170 @@
...
+	for _, o := range opts {
+		if err := o(cfg); err != nil {
+			return nil, err
+		}
+	}
+	i, err := complementIdentity(cfg.identity)
+	if err != nil {
+		return nil, err
+	}
+	cfg.identity = *i
```
Benehiko:
```suggestion
	cfg.identity = i
```

we shouldn't need to de-reference

pkg/stub/config.go
```diff
@@ -0,0 +1,170 @@
...
+func WithConnection(conn net.Conn) Option {
+	return func(s *cfg) error {
+		s.conn = conn
+		return nil
+	}
+}
+
+type cfg struct {
+	plugin Plugin
+	identity
```
Benehiko:
```suggestion
	*identity
```

pkg/stub/config.go
joe0BAB: that's the thing here: They are not for manual use but for being used by the engine. So when any one of them is set, we assume the engine is launching the plugin (we then later check if they actually all are set and error if not).

joe0BAB: I tried to make this explicit, but maybe we need to make it more explicit?

pkg/stub/config.go
joe0BAB: Why? Then I have to deal with nil pointers everywhere. I'm not very keen to do that. Also this is mandatory, ie., if we can't get this setup, we're returning an error anyway.

pkg/stub/config.go
joe0BAB: Nice! Will do.

pkg/stub/config.go
Benehiko: it's just that you de-reference everywhere, which makes a copy.

joe0BAB: I see. But in the code that will actually use it later, not having a reference will be much easier/safer (I don't have it pushed yet)

pkg/stub/stub.go
joe0BAB: Fix logging is on my todo list for later as well. Just wanted to have something so I can place log statements. But don't like it much either.

pkg/stub/config.go
Benehiko: yeah this is not very explicit

pkg/stub/stub.go
Benehiko: zap looks pretty good
https://github.com/uber-go/zap

pkg/adaptation/adaptation.go
joe0BAB: Actually had to address it now because it was already hitting me in the unit tests.

pkg/stub/config.go
```diff
@@ -0,0 +1,226 @@
...
+	conn                net.Conn
+	registrationTimeout time.Duration
+}
+
+func newCfg(p Plugin, opts ...ManualLaunchOption) (*cfg, error) {
+	if ShouldHaveBeenLaunchedByEngine() {
+		logrus.Info("Plugin launched by engine, restoring config...")
+		if len(opts) > 0 {
+			return nil, errors.New("plugin launched by secrets engine, cannot use manual launch options")
```
joe0BAB: This is a bit up for debate. I'm not sure what's best here, should we return an error or not?

Alternatively we also could move this into the user code so people have to explicitly write an if using `ShouldHaveBeenLaunchedByEngine()` helper and then we had to exported New functions, one with options one without. But I don't like that either, it feels like to much boilerplate.

joe0BAB: Or we could just not return an error and silently ignore options when launched by the engine.

pkg/stub/config.go
```diff
@@ -0,0 +1,226 @@
...
+// Note: There's no 100% guarantee that this really happened, but we take the custom internal
+// environment variables as indication that it should have happened.
+func ShouldHaveBeenLaunchedByEngine() bool {
+	// In theory, all variables should always be set by the engine and we'd just need to check one.
+	// But there could be a bug, so we check for any and verify all values are set correctly later.
+	name := os.Getenv(adaptation.PluginNameEnvVar)
+	idx := os.Getenv(adaptation.PluginIdxEnvVar)
+	timeoutStr := os.Getenv(adaptation.PluginRegistrationTimeoutEnvVar)
+	env := os.Getenv(adaptation.PluginSocketEnvVar)
+	return name != "" || idx != "" || timeoutStr != "" || env != ""
```
Benehiko: this looks better with the function name, but still unsure why we need to wait until later to error when not set up correctly.

Benehiko: we could actually just call `restoreConfig` here and return `(configFromEngine, error)`. 

The function also doesn't need to be exported I think, since the plugin doesn't need to know about the engine internals I guess? We fallback to manual mode always and could log out an error that "engine mode" is disabled for this plugin. If manual mode also fails, then just simply error and return.

pkg/stub/config.go
Benehiko: should the user be in control of this? I would think that if the plugin exists inside a PATH or directory we are aware of then I guess it should always be picked up? Perhaps that needs more discussion (in the cases of disabling a plugin, overrides etc.)

pkg/stub/config.go
```diff
@@ -0,0 +1,131 @@
...
+		conn, err := net.Dial("unix", defaultSocketPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to connect to default socket %q: %w", defaultSocketPath, err)
+		}
+		cfg.conn = conn
+	}
+	if cfg.name == "" {
+		if len(os.Args) == 0 {
+			// This should never happen in practice but can happen in tests or when something else empties os.Args for whatever reason.
+			return nil, errors.New("plugin name must be specified (could not derive from os.Args)")
```
joe0BAB, line 85: This is a bit unfortunate. But I believe eg `go test` does it (that made the tests crash on CI so I noticed). Otherwise we could have moved that up to the line where the struct gets created initially.

pkg/stub/config.go
```diff
@@ -0,0 +1,131 @@
...
+
+func newCfg(p Plugin, opts ...ManualLaunchOption) (*cfg, error) {
+	engineCfg, err := restoreConfig(p)
+	if err != nil && !errors.Is(err, errPluginNotLaunchedByEngine) {
+		return nil, err
+	}
+	if err == nil && engineCfg != nil {
+		return engineCfg, nil
+	}
+	return newCfgForManualLaunch(p, opts...)
```
Benehiko:
```suggestion
	if errors.Is(err, errPluginNotLaunchedByEngine) {
		return newCfgForManualLaunch(p, opts...)
	}
	return engineCfg, err
```

Shorten this and make the `manual` mode a fallback instead of the _default case_

pkg/adaptation/adaptation.go
```diff
@@ -0,0 +1,59 @@
...
+	var result PluginConfigFromEngine
+	if err := json.Unmarshal([]byte(in), &result); err != nil {
+		return nil, fmt.Errorf("failed to decode plugin config from engine %q: %w", PluginLaunchedByEngineVar, err)
+	}
+	if result.Name == "" {
+		return nil, errors.New("plugin name is required")
+	}
+	if result.RegistrationTimeout == 0 {
+		return nil, errors.New("plugin registration timeout is required")
+	}
```
Benehiko, line 53: Wonder if it's necessary to error on a timeout value, perhaps just fallback to a sane default instead?

pkg/adaptation/adaptation.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (c *PluginConfigFromEngine) ToString() (string, error) {
+	result, err := json.Marshal(c)
+	if err != nil {
+		return "", err
+	}
+	return string(result), nil
+}
+
+func NewPluginConfigFromEngineFromString(in string) (*PluginConfigFromEngine, error) {
```
Benehiko:
```suggestion
func NewPluginConfigFromEngineEnv(envConfig string) (*PluginConfigFromEngine, error) {
```

Benehiko: the `FromEngineFromString` part threw me off a bit

pkg/adaptation/adaptation.go
```diff
@@ -0,0 +1,59 @@
...
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+const (
+	// PluginLaunchedByEngineVar is used to inform engine-launched plugins about their name.
+	PluginLaunchedByEngineVar = "DOCKER_SECRETS_ENGINE_LAUNCH_CFG"
```
Benehiko:
```suggestion
	PluginLaunchedByEngineVar = "DOCKER_SECRETS_ENGINE_PLUGIN_LAUNCH_CFG"
```

Sorry for being so nit picky, perhaps specifying _who this config is for_ as the environment name makes sense?

pkg/adaptation/adaptation_test.go
```diff
@@ -0,0 +1,88 @@
...
+	var sb strings.Builder
+	sb.Grow(n)
+	for i := 0; i < n; i++ {
+		if i%2 == 0 {
+			sb.WriteByte('a')
+		} else {
+			sb.WriteByte('b')
+		}
+	}
+	return sb.String()
```
Benehiko:
```suggestion
return strings.Repeat("ab", n)
```
https://pkg.go.dev/strings#Repeat

pkg/stub/config_test.go
```diff
@@ -0,0 +1,155 @@
...
+			test: func(t *testing.T) {
+				_, err := restoreConfig(mockPlugin{})
+				assert.ErrorIs(t, err, errPluginNotLaunchedByEngine)
+			},
+		},
+		{
+			name: "invalid config from the engine",
+			test: func(t *testing.T) {
+				defer cleanupEnv()
+				_ = os.Setenv(adaptation.PluginLaunchedByEngineVar, "test-plugin")
```
Benehiko:
```suggestion
				t.SetEnv(adaptation.PluginLaunchedByEngineVar, "test-plugin")
```

`t.SetEnv` cleans up environment variables for you.

pkg/stub/config_test.go
```diff
@@ -0,0 +1,155 @@
...
+				_, err := restoreConfig(mockPlugin{})
+				assert.Error(t, err)
+			},
+		},
+		{
+			name: "valid config",
+			test: func(t *testing.T) {
+				sockets, err := nriNet.NewSocketPair()
+				require.NoError(t, err)
+				defer sockets.Close()
```
Benehiko: You should probably use `t.Cleanup` instead of defer. The behavior is very similar, but there are some non-obvious pitfalls with `defer` https://github.com/docker/cli/pull/5645#discussion_r1860912611

Benehiko: would also be good to mark the tests as parallel (should be able to work right?)

joe0BAB: Parallel doesn't work unfortuantelly: `panic: testing: test using t.Setenv or t.Chdir can not use t.Parallel`

(And switching to using `t.Setenv` actually caught it, which is very good!

pkg/adaptation/adaptation.go
joe0BAB: Let's keep for now. In NRI they didn't error on this which made them miss that in NRI configuring the registration timeout is actually bugged / doesn't work.

pkg/stub/config.go
```diff
@@ -0,0 +1,128 @@
...
+	conn                net.Conn
+	registrationTimeout time.Duration
+}
+
+func newCfg(p Plugin, opts ...ManualLaunchOption) (*cfg, error) {
+	engineCfg, err := restoreConfig(p)
+	if errors.Is(err, errPluginNotLaunchedByEngine) {
+		return newCfgForManualLaunch(p, opts...)
+	}
+	return engineCfg, nil
```
Benehiko: You'll need to return the error as well (it might be set).

```suggestion
	return engineCfg, err
```

Benehiko (commented):
One small change, then we're good to go!

pkg/stub/config.go
joe0BAB: Thanks for catching that!!

Benehiko (approved):

PR #22, opened 2025-06-20
feat(sdk): stub ipc

joe0BAB:


pkg/stub/ipc.go
```diff
@@ -0,0 +1,91 @@
...
+	teardown  func() error
+}
+
+type Ipc interface {
+	conn() net.Conn
+	wait(ctx context.Context) error
+	close()
+}
+
+func NewIPC(sockConn net.Conn, handler http.Handler) (result Ipc, retErr error) {
```
joe0BAB: This export is just temporary to make the linter happy. Will remove once it's used in follow up PR / on the stub.

pkg/stub/ipc.go
```diff
@@ -0,0 +1,91 @@
...
+	mux := multiplex.Multiplex(sockConn)
+	defer func() {
+		if retErr != nil {
+			mux.Close()
+		}
+	}()
```
Benehiko: I don't quite understand this, wouldn't we _always_ want to close the multiplex - even if something errored?

pkg/stub/ipc.go
```diff
@@ -0,0 +1,91 @@
...
+	doneC := make(chan struct{})
+	listener, err := mux.Listen(multiplex.PluginServiceConn)
+	if err != nil {
+		mux.Close()
```
Benehiko, line 53: we already registered a `defer mux.Close()` at line 30, just remove the if check and the behavior will be the same

pkg/stub/ipc.go
```diff
@@ -0,0 +1,91 @@
...
+	}()
+	return &ipcImpl{
+		mConn:     conn,
+		doneC:     doneC,
+		serverErr: serverErr,
+		teardown: sync.OnceValue(func() error {
+			_ = listener.Close()
+			_ = server.Close()
+			_ = conn.Close()
+			_ = mux.Close()
```
Benehiko: Perhaps we can `errors.Join`?

pkg/stub/ipc.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}()
+	conn, err := mux.Open(multiplex.RuntimeServiceConn)
+	if err != nil {
+		return nil, fmt.Errorf("failed to multiplex grcp client connection: %w", err)
+	}
+	defer func() {
+		if retErr != nil {
+			_ = conn.Close()
+		}
```
Benehiko: I think we want to always close right?

pkg/stub/ipc.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	var serverErr error
+	go func() {
+		serverErr = server.Serve(listener)
+		close(doneC)
+	}()
+	defer func() {
+		if retErr != nil {
+			server.Close()
+		}
```
Benehiko: same here

pkg/stub/ipc.go
```diff
@@ -0,0 +1,91 @@
...
+	case <-i.doneC:
+		return i.serverErr
+	case <-ctx.Done():
+		return nil
+	}
+}
+
+func (i *ipcImpl) close() {
+	_ = i.teardown()
+}
```
Benehiko:
```suggestion
func (i *ipcImpl) close() error {
	return i.teardown()
}
```

pkg/stub/ipc.go
```diff
@@ -0,0 +1,91 @@
...
+		mux.Close()
+		return nil, err
+	}
+	server := &http.Server{
+		Handler: handler,
+	}
+	var serverErr error
+	go func() {
+		serverErr = server.Serve(listener)
+		close(doneC)
```
Benehiko: We should check if the returned error from `server.Serve` is not just an `http.ErrServerClosed`.
https://github.com/docker/hub-core-sdk/blob/main/server/server.go#L111-L119

Benehiko (commented):
Yeah this code is hard to read, it's doing a lot of things which aren't too obvious.

I'll have to think a bit more about how we can separate the implementation details

pkg/stub/ipc.go
joe0BAB: this should have been `listener` instead of `mux`

pkg/stub/ipc.go
```diff
@@ -0,0 +1,94 @@
...
+		mConn:  conn,
+		server: server,
+		teardown: sync.OnceValue(func() error {
+			var errList []error
+			if err := server.server.Close(); err != nil {
+				errList = append(errList, err)
+			}
+			if err := mux.Close(); err != nil {
+				errList = append(errList, err)
+			}
```
Benehiko:
```suggestion
			err := errors.Join(server.server.Close(), mux.Close())
```

The if's are unnecessary, and any nil error would be dropped by `errors.Join`.

Benehiko (commented):
Looks a lot better!
Just a small nit

Benehiko (approved):

PR #24, opened 2025-06-20
Add store module

Benehiko:
The secrets engine is opinionated on the key format (id of the secret). Each secret stored should store or at a minimum map the key to the `secrets.ID`. 

This patch adds two new interfaces to the secrets engine:
- `Store`
- `Secret`

A `Store` can be a physical file or a wrapper around an `API`.
A `Secret` can be anything as long as it can be `Marshalled` and `Unmarshalled`.

The implementer of a `Secret` is responsible for maintaining the format of a secret. The secrets engine does not enforce a particular encoder or marshaller. The implementer is free to decide.


store/store.go
```diff
@@ -0,0 +1,55 @@
...
+	Marshal() ([]byte, error)
+	// Unmarshal the secret from a slice of bytes into its structured format
+	Unmarshal(data []byte) error
+}
+
+// Store defines a strict format for secrets to conform to when interacting
+// with the secrets engine
+type Store interface {
+	// Erase removes credentials from the store for a given ID.
+	Erase(ctx context.Context, id ID) error
```
joe0BAB: `Delete`? I'd stick to conventions which usually is `Delete` or `Remove`.

store/store.go
```diff
@@ -0,0 +1,55 @@
...
+	// Get retrieves credentials from the store for a given ID.
+	Get(ctx context.Context, id ID) (Secret, error)
+	// GetAll retrieves all the credentials from the store.
+	GetAll(ctx context.Context) (map[ID]Secret, error)
+	// Store saves credentials in the store.
+	Store(ctx context.Context, id ID, secret Secret) error
```
joe0BAB: `Set` or `Save`? Same reasons as above.

joe0BAB (approved):

PR #25, opened 2025-06-20
store: add keychain package

Benehiko:
Required by https://github.com/docker/secrets-engine/pull/16

This patch adds the `keychain` package inside the `store` module, but does not implement an actual underlying keychain provider.


store/keychain/keychain.go
```diff
@@ -0,0 +1,53 @@
...
+			return errors.New("the prefix cannot be empty")
+		}
+		ks.keyPrefix = prefix
+		return nil
+	}
+}
+
+// New creates a new keychain store
+//
+// collectionID is a singular noun indicating the collection name, e.g. "docker"
```
joe0BAB: I don't get this comment. Maybe you moved code around?

store/keychain/keychain_dummy.go
```diff
@@ -0,0 +1,29 @@
...
+	"context"
+
+	"github.com/docker/secrets-engine/store"
+)
+
+var _ store.Store = &keychainStore[store.Secret]{}
+
+// Erase implements secrets.Store.
+func (k *keychainStore[T]) Delete(ctx context.Context, id store.ID) error {
+	panic("unimplemented")
```
joe0BAB, line 13: I don't get the point of these. Is it just documentation purpose? Or do you tend to implement them later?

I don't see much value in the code as is, but maybe I'm missing something.

store/mocks/mock_store.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"maps"
+	"sync"
+
+	"github.com/docker/secrets-engine/store"
+)
+
+type MockStore struct {
+	lock  sync.RWMutex
```
joe0BAB, line 12: nit: `m` (at least in piñata we always name this `m`)

store/mocks/mock_store.go
```diff
@@ -0,0 +1,65 @@
...
+	store map[store.ID]store.Secret
+}
+
+func (m *MockStore) init() {
```
joe0BAB, line 16: nit: `s` (if we'd go with renaming `lock` to `m`)

joe0BAB, line 16: Why not have a `new` function instead?

store/keychain/keychain.go
Benehiko: oof, yeah i forgot to change the docs

store/keychain/keychain_dummy.go
Benehiko: it's to satisfy the `store.Store` interface. The actual implementation has been done somewhere else. I just wanted to break up some of the PR.

store/mocks/mock_store.go
Benehiko: I'm not so sure I understand why `m` is better than `lock`, sounds too obscure.

store/mocks/mock_store.go
Benehiko: `new` is a keyword in Go

store/mocks/mock_store.go
Benehiko: searching through [containerd/containerd](https://github.com/search?q=repo%3Acontainerd%2Fcontainerd%20sync.RWMutex&type=code) I see a lot of `lock` being used. Other variations are also used: `mu`, `m`, `l` and `rwlock`.

store/mocks/mock_store.go
joe0BAB: right, I meant `newSomething`, eg `newMockStore`

Benehiko: I can rename it to `tryInit` instead. `newSomething` sounds like it will return a new object

joe0BAB (approved):
LGTM! (I'd still prefer not having the `init` function on the mock, but it's not blocking to me)

PR #26, opened 2025-06-20
Secrets service experiment

Benehiko:
This is an experiment - it is a simple package that tests the `org.freedesktop.secrets` API over DBUS. It tries to list and create collections.


Benehiko:
this is already supported in main. Will close this.

PR #29, opened 2025-06-23
feat(store): encrypted filestore

Benehiko:
This commit introduces an encrypted filestore backed by the os.Root type,
which enforces a flat storage structure confined to a specific directory.
Each secret is stored per encryption function in its own file, alongside a
public metadata file in JSON format. Filenames are base64-encoded secret
IDs to avoid issues with forward slashes ("/") and other special characters.

All secret files are encrypted with the filippo.io/age library, which
supports multiple encryption and decryption methods, including hardware
keys, SSH keys, and custom plugins. The store cannot follow symlinks
outside the directory specified by os.Root, ensuring stronger isolation.


store/filestore/filestore.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+	return data, nil
+}
+
+func getFile(ctx context.Context, filesystem *os.Root, secretFileName, metadataFileName string) (*file, error) {
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
```
joe0BAB: This smells. Seems like we don't use the context anywhere else in this function. Maybe remove this code and the context argument?

store/filestore/filestore.go
```diff
@@ -0,0 +1,375 @@
...
+var _ store.Store = &fileStore[store.Secret]{}
+
+type file struct {
+	metadata        map[string]string
+	encryptedSecret []byte
+}
+
+// getFileNames retrieves the base64 encoded string of the secret ID and
+// metadata file
+func getFileNames(id store.ID) (string, string) {
```
joe0BAB:
```suggestion
// toFileNames derives a base64 encoded string from the secret ID and
// creates a corresponding metadata file name
func toFileNames(id store.ID) (string, string) {
```

(it's a conversion, not a "get")

store/filestore/filestore.go
```diff
@@ -0,0 +1,375 @@
...
+		encryptedSecret: encryptedSecret,
+	}, nil
+}
+
+func saveFile(ctx context.Context, id store.ID, f file, filesystem *os.Root) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+	}
```
joe0BAB: why do we need this?

store/go.mod
```diff
@@ -1,13 +1,14 @@
 module github.com/docker/secrets-engine/store
 
-go 1.24.3
+go 1.25
```
joe0BAB: This change smells being part of this PR. I'm fine with keeping it here, but do we need an issue that this inconsistency should have been detected by our CI?

Benehiko: this PR requires 1.25. I'll make another PR to bump the versions quick

Benehiko: https://github.com/docker/secrets-engine/pull/269

store/posixage/filestore.go
```diff
@@ -0,0 +1,333 @@
...
+	return &file{
+		metadata:        metadata,
+		encryptedSecret: encryptedSecret,
+	}, nil
+}
+
+func saveFile(_ context.Context, id store.ID, f file, filesystem *os.Root) error {
+	secretFileName, metadataFileName := toFileNames(id)
+	tmpSecretFileName := secretFileName + ".tmp"
+	tmpMetadataFileName := metadataFileName + ".tmp"
```
joe0BAB: nit: maybe this needs a comment explaining why we do this extra work

store/posixage/filestore.go
```diff
@@ -0,0 +1,333 @@
...
+	if err := tmpSecret.Sync(); err != nil {
+		return err
+	}
+
+	if err := filesystem.Rename(tmpMetadataFileName, metadataFileName); err != nil {
+		return err
+	}
+	if err := filesystem.Rename(tmpSecretFileName, secretFileName); err != nil {
+		return err
+	}
```
joe0BAB: this needs to be in reverse order: the metadata file is the primary file. When it becomes "visible" to the posix file storage, the other file already needs to have been renamed (so this behaves atomic) -> ideally, we also have a test for this

store/posixage/filestore.go
joe0BAB: There's a concurrency issue here: let's say we have to consecutive `saveFile` calls for the same `store.ID`. Then we have no guarantee about the outcome. It could be the first or second that "wins", but it could also be a mix, metadata from the first with the value from the second. The issue is that `filesystem.Create` always overwrites. We could fix this by adding some random string when creating the files.

store/posixage/filestore.go
```diff
@@ -0,0 +1,333 @@
...
+
+	if _, err = tmpSecret.Write(f.encryptedSecret); err != nil {
+		return err
+	}
+	if err := tmpSecret.Sync(); err != nil {
+		return err
+	}
+
+	if err := filesystem.Rename(tmpMetadataFileName, metadataFileName); err != nil {
+		return err
```
joe0BAB: Would it make sense to wrap `os.IsExist(err)` into a dedicated error type and document it? I'm wondering with the current implementation how explicit the behavior is that a `saveFile` would fail with a `os.IsExist` (and that to solve it, the existing needs to be deleted, ie, that there's no overwrite option with the current implementation)

store/posixage/filestore_test.go
```diff
@@ -0,0 +1,179 @@
...
+		assert.Len(t, fsFiles, 4)
+
+		storeSecrets, err := s.Filter(t.Context(), store.MustParsePattern("something/*"))
+		require.NoError(t, err)
+
+		id := store.MustParseID("something/secret1")
+		assert.Len(t, storeSecrets, 1)
+		assert.IsType(t, &mocks.MockCredential{}, storeSecrets[id])
+		assert.EqualValues(t, secrets[id], storeSecrets[id])
+	})
```
joe0BAB: We need a few more things to be verified through tests:
* there's defined behavior for concurrent writes (ie concurrent writes are guaranteed to not result in an undefined state)
* there's defined behavior for concurrent delete/write operations (ie concurrent delete/write are guaranteed to not result in an undefined state)

store/posixage/filestore.go
```diff
@@ -0,0 +1,333 @@
...
+		return nil, err
+	}
+
+	return &file{
+		metadata:        metadata,
+		encryptedSecret: encryptedSecret,
+	}, nil
+}
+
+func saveFile(_ context.Context, id store.ID, f file, filesystem *os.Root) error {
```
joe0BAB: What about cleanup logic on early returns? Eg, wouldn't we want to delete our temp files when the renaming fails later?

store/posixage/filestore.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+func (f *fileStore[T]) Delete(_ context.Context, id store.ID) error {
+	secretsFileName, metadataFileName := toFileNames(id)
+	if err := f.filesystem.Remove(metadataFileName); err != nil {
+		return err
+	}
+	if err := f.filesystem.Remove(secretsFileName); err != nil {
+		return err
+	}
```
joe0BAB: This should be in a transaction. As far as I'm aware, wrapping things in a directory would be the way to do this.

store/posixage/store.go
```diff
@@ -0,0 +1,566 @@
...
+}
+
+var _ store.Store = &fileStore[store.Secret]{}
+
+type secretFile struct {
+	fileName      string
+	encryptedData []byte
+}
+
+type file struct {
```
joe0BAB: this is not a file anymore but a folder with files

Maybe rename to `posixSecret`? `fsSecret`?

store/posixage/store.go
```diff
@@ -0,0 +1,566 @@
...
+	return nil, errors.New("could not decrypt secret with provided keys")
+}
+
+func (f *fileStore[T]) Delete(_ context.Context, id store.ID) error {
+	f.l.Lock()
+	defer f.l.Unlock()
+
+	encFile := file{
+		rootName: base64.StdEncoding.EncodeToString([]byte(id.String())),
+	}
```
joe0BAB: Wrap this into `fromID(id store.Secret) file`? I see this repeated a lot through this file. Plus it would make this function a 1-liner.

store/posixage/store.go
```diff
@@ -0,0 +1,566 @@
...
+	"github.com/docker/secrets-engine/store"
+	"github.com/docker/secrets-engine/x/logging"
+)
+
+type fileStore[T store.Secret] struct {
+	filesystem                *os.Root
+	registeredEncryptionFuncs []callbackFunc
+	registeredDecryptionFunc  []callbackFunc
+	f                         store.Factory[T]
+	l                         sync.RWMutex
```
joe0BAB: This smells. While it's true that it solves concurrency within one store process, it doesn't solve concurrency of multiple store processes operating on the same physical files (worse, it might give the illusion there is concurrency safety). We should remove this and either use FS locks (but they are OS dependent, so we should try hard to avoid them) or solve this differently, ideally with a better architecture/design that leaves that to the OS/FS. (FUSE could also be an option, but implementing a FUSE is OS specific as well, so then probably FS locks would be the lesser evil)

store/posixage/store.go
```diff
@@ -0,0 +1,642 @@
...
+
+		var encryptedSecret bytes.Buffer
+		w, err := age.Encrypt(&encryptedSecret, recipients...)
+		if err != nil {
+			return err
+		}
+		defer func() {
+			_ = w.Close()
+		}()
+
```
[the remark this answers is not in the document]
Benehiko: no

store/posixage/flock.go
```diff
@@ -0,0 +1,111 @@
...
+	fl, err := openFile(root)
+	if err != nil {
+		return nil, err
+	}
+
+	ep := backoff.NewExponentialBackOff()
+	ep.InitialInterval = time.Millisecond * 10
+	ep.MaxInterval = time.Millisecond * 100
+
+	unlock, err := backoff.Retry(context.Background(), func() (unlockFunc, error) {
```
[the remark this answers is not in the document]
Benehiko: no

joe0BAB: From looking at where `attemptLock` is used, we should be able to easily pass the appropriate context through.

store/posixage/flock.go
```diff
@@ -0,0 +1,111 @@
...
+			// we want the underlying error - something went wrong to recover
+			// the file (e.g. could not remove).
+			return nil, errors.Join(err, recoverErr)
+		}
+		if recoverErr != nil && errors.Is(recoverErr, errRecoverLock) {
+			return nil, err
+		}
+
+		// we could recover, re-run the lock
+		return attemptLock(root, exclusive, timeout)
```
joe0BAB: Passing in the same timeout here smell for an invite loop: The backoff.Retry above errors out due to time out an we re-run it here again with the same timeout..

I believe what we want to use instead is something like a deadline (eg context.WithDeadline might help).

store/posixage/flock.go
```diff
@@ -0,0 +1,111 @@
...
+
+	ep := backoff.NewExponentialBackOff()
+	ep.InitialInterval = time.Millisecond * 10
+	ep.MaxInterval = time.Millisecond * 100
+
+	unlock, err := backoff.Retry(context.Background(), func() (unlockFunc, error) {
+		return lockFile(fl, exclusive)
+	},
+		backoff.WithBackOff(ep),
+		backoff.WithMaxElapsedTime(timeout),
```
joe0BAB: ~Couldn't we just use the context for this? Something smells here.~

store/posixage/flock.go
```diff
@@ -0,0 +1,111 @@
...
+// attemptLock acquires a file-based lock using a temporary file.
+//
+// If the lock file does not exist, it is created. If the lock cannot be
+// acquired immediately, the function retries until the specified timeout
+// is reached.
+//
+// To recover from stale locks (e.g., if a process crashes before releasing),
+// the function removes the lock file if its modification time is older than
+// 30 seconds. This assumes no valid process would hold the lock that long.
+func attemptLock(root *os.Root, exclusive bool, timeout time.Duration) (unlockFunc, error) {
```
joe0BAB:
```suggestion
func attemptLock(ctx context.Context, root *os.Root, exclusive bool) (unlockFunc, error) {
```

I think we should just re-use contexts for this and leave it up to the caller when to abort `attemptLock` (eg this is how they do it in `http.Server.Shutdown` as well, that they have no timeout argument but instead you need to set a timeout on the context).

store/posixage/flock.go
joe0BAB: There are two kinds of timeouts in this function: The overall timeout how long we recursively run `attemptLock` and the timeout until we give up per `attemptLock` and go into recovery. The first should be handled through a context from the caller. I'm not sure the "timeout until we go into recovery" even should be configurable, sounds more like every consumer of that `attemptLock` should adhere to the same rules for recovery.

store/posixage/flock.go
```diff
@@ -0,0 +1,111 @@
...
+		// recoverLock closes fl
+		recoverErr := recoverLock(root, fl)
+		if recoverErr != nil && !errors.Is(recoverErr, errRecoverLock) {
+			// we want the underlying error - something went wrong to recover
+			// the file (e.g. could not remove).
+			return nil, errors.Join(err, recoverErr)
+		}
+		if recoverErr != nil && errors.Is(recoverErr, errRecoverLock) {
+			return nil, err
+		}
```
joe0BAB: This is a bit hard to read as is. We could go with the "filtering out a specific error" pattern that we already use at several other places in our codebase:
```suggestion
		if recoverErr := recoverLock(root, fl); recoverErr != nil{
			return nil, errors.Join(err, filterRecoverErr(recoverErr))
		}
```

and define
```
func filterRecoverErr(err error) error {
	if errors.Is(err, errRecoverLock) {
		return nil
	}
	return err
}
```

store/posixage/flock.go
```diff
@@ -0,0 +1,111 @@
...
+
+		// we could recover, re-run the lock
+		return attemptLock(root, exclusive, timeout)
+	}
+	// truncate to update the modtime to signal to other processes that the
+	// current lock is valid so they don't attempt a recovery on it.
+	_ = fl.Truncate(0)
+
+	return unlock, nil
```
joe0BAB: So this comment might be a bit of personal style/preference.

I think it would improve readability to move the `sync.OnceValue` up here and just have to simple functions `lockFile(f *os.File, exclusive bool) error` and `unlockFile(f *os.File, exclusive bool) error`. It always feels like a risk/smell to me to distribute logic that belongs together across multiple functions. That decoupling would also allow to wrap the backoff stuff into a `tryLockFile(cox context.Context, f *os.File, exclusive bool) error` which would further allow to focus on the logic in `attemptLock`.

store/posixage/flock.go
joe0BAB: Could we rewrite this from recursion to iteratively running `tryLock`/`recoverLock`?

store/posixage/flock_unix.go
```diff
@@ -0,0 +1,42 @@
...
+	} else {
+		flag |= unix.LOCK_SH
+	}
+
+	err := unix.Flock(int(f.Fd()), flag)
+	if err != nil {
+		return nil, errors.Join(ErrLockUnsuccessful, err)
+	}
+
+	return sync.OnceValue(func() error {
```
joe0BAB: nit: move this up to `attemptLock` where all the logic is and have a separate "dumb" function `unlockFile`

store/posixage/flock_test.go
```diff
@@ -0,0 +1,128 @@
...
+		// change the lock file modification time
+		fakeModTime := time.Date(2020, 1, 1, 0, 0, 0, 0, time.UTC)
+		require.NoError(t, root.Chtimes(lockFileName, fakeModTime, fakeModTime))
+
+		unlock, err := attemptLock(root, exclusive, time.Millisecond)
+		require.NoError(t, err)
+		require.NoError(t, unlock())
+	})
+
+	t.Run("recoverLock errors if a recover was not possible", func(t *testing.T) {
```
joe0BAB: Got a bit confused by this title but the test itself makes sense. Maybe something like this could help?
```suggestion
	t.Run("recoverLock does not recover if the lock is newer than 30sec but errors instead", func(t *testing.T) {
```
(to me it wasn't clear that "not possible" means the 30sec haven't expired yet)

joe0BAB: nit: we could group this and the other function that tests recoverLock into a separate test

store/posixage/flock.go
Benehiko: timeout is max time spent retrying - infinite loop is only possible if:
1. lock was unsuccessful
2. recovery was successful

So something else is wrong in getting a lock (e.g. some odd OS behavior). It cannot be permission errors, if it was then recovery would also fail (it removes the file).

To put your mind at ease with the recursive function, i'll refactor it so that it attempts one last time at locking and instead just returns on error.

store/posixage/flock.go
Benehiko: i'm not sure what you mean - timeout is the max time it will wait for the lock to succeed. If it does not succeed it retries by going into recovery + retry again with timeout. `attemptLock` has no timeout, meaning (there is no global timeout for the function first try + recovery + second try).

Benehiko: Currently there are two things we should address:
1. how long we should delay between retries (with exponential backoff it increases over time)
2. how long in total we should keep trying a lock (currently this is specified by timeout).

We could use ctx for 2. and rely solely on that for "how long we should keep trying a lock".

store/posixage/store.go
```diff
@@ -0,0 +1,642 @@
...
+	"filippo.io/age"
+
+	"github.com/docker/secrets-engine/store"
+	"github.com/docker/secrets-engine/x/logging"
+)
+
+type fileStore[T store.Secret] struct {
+	filesystem *os.Root
+	factory    store.Factory[T]
+	l          sync.RWMutex
```
joe0BAB, line 34: Do we still need that? I thought we replaced it with a file based lock/mutex.

store/posixage/store.go
```diff
@@ -0,0 +1,642 @@
...
+			_ = w.Close()
+		}()
+
+		if _, err := w.Write(val); err != nil {
+			return err
+		}
+
+		if err := w.Close(); err != nil {
+			return err
+		}
```
joe0BAB, line 354: Do we need to call close here? We already deferred a close call a few lines above.

store/posixage/store.go
```diff
@@ -0,0 +1,642 @@
...
+}
+
+var _ store.Store = &fileStore[store.Secret]{}
+
+type secretFile struct {
+	fileName      string
+	encryptedData []byte
+}
+
+type secretDirectory struct {
```
joe0BAB: 1) Could we add `fs *os.Root` as a field here so that we'd have `newSecretDirectory(fs *os.Root, id store.ID) *secretDirectory`? We already pass this in on any function of this struct.

2) Could we move this into a separate file with its own set of tests? (eg `sdir.go`+`stir_test.go` or something)

(also the methods of this struct are a good candidate to introduce an interface and allow mocking for tests at some point)

joe0BAB: (I'd just write a simple "roundtrip tests" that verifies the save/restore cycle)

store/posixage/store.go
```diff
@@ -0,0 +1,642 @@
...
+	// rootName is a base64 encoding of the secret ID
+	rootName string
+	metadata map[string]string
+	secrets  []secretFile
```
joe0BAB: these fields are a smell here, the restore function(s) should just return those values and they should be parameters to the save function

store/posixage/store.go
```diff
@@ -0,0 +1,642 @@
...
+		secrets = append(secrets, secretFile{
+			fileName:      fileName,
+			encryptedData: encryptedSecret.Bytes(),
+		})
+	}
+
+	encFile := newSecretDirectory(id)
+	encFile.metadata = metadata
+	encFile.secrets = secrets
+	return encFile.save(f.filesystem)
```
joe0BAB: nah bro, let's just turn this into a proper function with arguments

joe0BAB:
```
sDir := newSecretDirectory(f.filesystem, id)
sDir.save(metadata, secrets)
```

store/posixage/store.go
```diff
@@ -0,0 +1,642 @@
...
+	if err != nil {
+		return err
+	}
+
+	var metadata map[string]string
+	if err := json.Unmarshal(b, &metadata); err != nil {
+		return err
+	}
+
+	f.metadata = metadata
```
joe0BAB:
```suggestion
```

(we already have a function for this right below this function)

store/posixage/store.go
```diff
@@ -0,0 +1,642 @@
...
+
+		encFile := &secretDirectory{
+			rootName: dir.Name(),
+		}
+
+		if err := encFile.restore(f.filesystem); err != nil {
+			continue
+		}
+
+		decryptedSecret, err := f.decryptSecret(ctx, encFile)
```
joe0BAB:
```suggestion
		sDir := fromDirName(f.filesystem, dir.Name())
		metadata, err:= sDir.ReadMetadata()
		if err != nil { ... }
		secrets, err := sDir.ReadSecrets()
		if err != nil { ... }
		decryptedSecret, err := f.decryptSecret(ctx, secrets)
```

(`secretDirectory` should be stateless / only keep a pointer to the underlaying FS as member variable, similar like the Go filesystem library functions)

store/posixage/store.go
```diff
@@ -0,0 +1,642 @@
...
+				secrets[secretsIdx], secrets[i] = secrets[i], secrets[secretsIdx]
+				secretsIdx++
+				break
+			}
+		}
+	}
+	return nil
+}
+
+func (f *fileStore[T]) decryptSecret(ctx context.Context, encSecret *secretDirectory) ([]byte, error) {
```
joe0BAB:
```suggestion
func (f *fileStore[T]) decryptSecret(ctx context.Context, secrets []secretFile) ([]byte, error) {
```

(no need to depend on `secretDirectory`)

store/posixage/store.go
Benehiko: yes, this forces the last bytes to be written. Doing it on defer is not optimal since we want to rename the file. calling close inside the defer is just precaution (e.g. we return early)

store/posixage/store.go
Benehiko: hmm, i'm thinking i should try make this simpler. the rootName is used to log info for when a decrypt fails.

store/posixage/prompt.go
```diff
@@ -0,0 +1,100 @@
...
+// to collect encryption keys, grouped by their [secretfile.KeyType].
+//
+// Each callback is called in order. The returned key is trimmed of whitespace
+// and validated to ensure it is not empty. Keys are then grouped into a map,
+// where the map key is the callback's key type and the value is a slice of
+// all collected keys for that type.
+//
+// It returns an error if any callback fails, if the key type cannot be
+// determined, or if a callback returns an empty key.
+func promptForEncryptionKeys(ctx context.Context, funcs []promptCaller) (map[secretfile.KeyType][]string, error) {
```
Benehiko, line 81: TODO: wondering if we could have a UX where the user is prompted -> encryption happens, repeats until all encryption functions were called. 

Currently it prompts in a loop collects all keys -> encrypts in a loop.

The error messages might not co-inside with whatever prompt was called last. This might cause confusing UX to the one implementing or using the finished product. A user entered a key format incorrectly (not expected by age) and only sees the error after the last prompt was made - we should show the error immediately after the prompt was made.

store/posixage/store.go
```diff
@@ -0,0 +1,441 @@
...
+// performing concurrent reads and writes, since acquiring a file lock can
+// take time-especially when recovering from a stale lock.
+//
+// It returns an unlock function that must be called to release the lock.
+func (f *fileStore[T]) tryLock(ctx context.Context) (func(), error) {
+	f.l.Lock()
+
+	unlock, err := flock.TryLock(ctx, f.filesystem)
+	if err != nil {
+		return nil, err
```
joe0BAB, line 55: we need to unlock `f.l.Unlock()` before

store/posixage/store.go
```diff
@@ -0,0 +1,441 @@
...
+// performing concurrent reads and writes, since acquiring a file lock can
+// take time-especially when recovering from a stale lock.
+//
+// It returns an unlock function that must be called to release the lock.
+func (f *fileStore[T]) tryRLock(ctx context.Context) (func(), error) {
+	f.l.RLock()
+
+	unlock, err := flock.TryRLock(ctx, f.filesystem)
+	if err != nil {
+		return nil, err
```
joe0BAB, line 81: we need to call `f.l.RUnlock()` before

store/posixage/store.go
```diff
@@ -0,0 +1,441 @@
...
+	}
+
+	secret, err := s.Marshal()
+	if err != nil {
+		return err
+	}
+	metadata := s.Metadata()
+
+	var secrets []secretfile.EncryptedSecret
+	// each encryption key must be be used within its group - age does not
```
joe0BAB:
```suggestion
	// each encryption key must be used within its group - age does not
```

nit: typo

store/posixage/store_test.go
```diff
@@ -0,0 +1,680 @@
...
+			return nil
+		}
+		if d.IsDir() {
+			dirs = append(dirs, d)
+		}
+		return err
+	})
+}
+
+func TestPOSIXAge(t *testing.T) {
```
joe0BAB, line 63: nit: from experience it can be handy to have the tests provisioned with a real logger (vs the current default noop logger)

store/posixage/internal/secretfile/secret.go
```diff
@@ -0,0 +1,198 @@
...
+// removed before writing, ensuring that secrets encrypted with different
+// keys cannot become inconsistent.
+//
+// Inside the directory, the function creates:
+//   - metadata.json — a JSON-encoded metadata file (always public)
+//   - secret<KeyType> — one encrypted secret file per key type
+//
+// If any step fails, the directory is removed to prevent partial or
+// inconsistent state. An error is returned in such cases.
+func Persist(id store.ID, root *os.Root, metadata map[string]string, secrets []EncryptedSecret) error {
```
joe0BAB, line 88: non-blocking (eg for follow up PR): some small test that verifies the combination of Persist/Restore works

(I know we implicitly cover this on the next layer above / in `store_test.go`, it's still helpful eg when troubleshooting/debugging to easily spot what works independently and what not)

store/posixage/store.go
joe0BAB: non-blocking: we could have a test for that (I'd still prefer if all the tryLock/tryRLock functions would have been implemented decoupled from `fileStore` that would just be a member of `fileStore`. Such decoupled design would allow for more testing / reduces the cognitive load on understanding `fileStore`)

joe0BAB (approved):
LGTM! ~(the missing unlock calls should be fixed before merging though)~ 

(was already fixed, just didn't see as I was logged out..)

PR #32, opened 2025-06-23
store/keychain: require a serviceName and serviceGroup

### LINKED ISSUE #5: Support OS keychain [macos, linux, windows]

## Why in Secrets Engine?

## Fallback store

## Prior Art

Benehiko:
The macOS keychain scopes items based on the application bundle ID, managed by XCode. The macOS keychain service also supports setting a Keychain group which shares keychain items between applications. On Linux any application can prompt the user for any credential stored inside the keychain. 

For us to support a wide range of use cases we need to require the `serviceName` and `serviceGroup`. This helps make credentials searchable via their attributes field (support by macOS and Linux).


store/keychain/keychain.go
```diff
@@ -6,47 +6,54 @@ import (
...
+// per application inside the collection.
+// Thus, adding a service group does not prevent other applications from
+// accessing the secret.
+//
+// A ServiceName is a unique name of the application storing credentials, it is
+// important to keep the service name unchanged once the service has stored credentials.
+// Changing the service name can be done, but would require migrating existing credentials.
+//
+// [Factory] is a function used to instantiate new secrets of type T.
+func New[T store.Secret](serviceGroup, serviceName string, factory Factory[T]) (store.Store, error) {
```
wmluke, line 42: Do we have any secrets stored in the wild that will need to be migrated?

wmluke: Minor nit: should/do we have functional tests to validate this behavior across platforms?

joe0BAB (approved):

store/keychain/keychain.go
Benehiko: >Do we have any secrets stored in the wild that will need to be migrated?

Not here, we aren't storing any secrets with the `keychain` library yet, so nothing to be migrated. But someone implementing this library would need to be made aware of it.

>Minor nit: should/do we have functional tests to validate this behavior across platforms?

Not yet, tests will come once we have a baseline.

PR #35, opened 2025-06-24
feat(sdk): registration service on the engine side

joe0BAB:
Plumping code between protobuf and the actual registration logic (to be implemented in follow up PR).


pkg/adaptation/registration.go
```diff
@@ -0,0 +1,64 @@
...
+
+type onRegisteredFunc func(ctx context.Context, cfg pluginCfgIn) error
+
+type RegisterService struct {
+	m            sync.Mutex
+	onRegistered onRegisteredFunc
+	sent         pluginCfgOut
+}
+
+func NewRegisterService(cfg pluginCfgOut, registeredFunc onRegisteredFunc) *RegisterService {
```
joe0BAB, line 40: Temporary export to make the linter happy. Will remove once it's being used.

Benehiko (approved):

PR #38, opened 2025-06-24
Cleanup "store" submodule: add linting and remove double vendoring

joe0BAB:


joe0BAB:
@Benehiko will continue this when the submodule is ready for it.

PR #40, opened 2025-06-24
Cleanup submodule setup

joe0BAB:


joe0BAB:
Let's not do this now.

PR #41, opened 2025-06-24
feat: setup go work

joe0BAB:
Adds go work related files and a CI job to check for any pending changes.


go.work
```diff
@@ -0,0 +1,6 @@
+go 1.24.3
+
+use (
+	.
```
Benehiko, line 4: could we add `store` already, or will we need to first get some scaffolding done?

Benehiko (approved):

go.work
joe0BAB: We could but it prompts some `go mod tidy` issues.

Benehiko: okay, then I'll do it inside another PR somewhere

PR #44, opened 2025-06-25
feat(sdk): adaptation setup logic

joe0BAB:
This implements the setup logic that happens between we got some connection and a plugin is running (either the engine created/started both or it was already there from something else, both could happen), and then we wire things up:
1. do the IPC multiplexing with the engine side handlers
2. wait for the registration call from the plugin
3. validate the registration

and if wiring up went well, we return 
- the (valid) parameters we received from the plugin
- the IPC/connection that can be used to talk to the plugin
- a close function that can be used to shut everything down


pkg/adaptation/setup.go
```diff
@@ -0,0 +1,87 @@
...
+
+type SetupResult struct {
+	conn  net.Conn
+	cfg   pluginCfgIn
+	close func() error
+}
+
+func Setup(conn net.Conn, isLaunchedByEngine bool, cfg pluginCfgOut, acceptPattern func(secrets.Pattern) error) (*SetupResult, error) {
+	chRegistrationResult := make(chan registrationResult, 1)
+	defer close(chRegistrationResult)
```
Benehiko: I'm so paranoid about this. I've run into so many problems where a channel was closed but there is another piece of code somewhere pushing to the channel and then it causes a panic.

We use `chRegistrationResult` inside the `newRegistrationLogic` function which stores it inside `registrationLogic`. We then try to push a value to the channel on line 105 of registration.go. Something just has to be a bit racy and it will panic.

https://go.dev/play/p/bjSRt8HsPS-

pkg/adaptation/setup.go
```diff
@@ -0,0 +1,87 @@
...
+	httpMux := http.NewServeMux()
+	httpMux.HandleFunc("/health", func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte("ok"))
+	})
+	httpMux.Handle(resolverv1connect.NewEngineServiceHandler(newRegisterService(newRegistrationLogic(&setupValidator{
```
Benehiko:
```suggestion
	httpMux.Handle(
	resolverv1connect.NewEngineServiceHandler(
	    newRegisterService(
	        newRegistrationLogic(
	            &setupValidator{
```

I'm not sure how to restructure this, but it was very difficult to know what was going on here.
Perhaps we could just split up the `registrationLogic` part from the `RegistrationService` part.

pkg/adaptation/setup.go
```diff
@@ -0,0 +1,87 @@
...
+	acceptPattern      func(secrets.Pattern) error
+}
+
+func (p setupValidator) Validate(in pluginCfgIn) (*pluginCfgOut, error) {
+	if p.isLaunchedByEngine && in.name != "" {
+		return nil, errors.New("plugin already registered with a different name")
+	}
+	if !p.isLaunchedByEngine && in.name == "" {
+		return nil, errors.New("plugin name is empty")
+	}
```
Benehiko: The validation checks here seem a bit strange to me. 

Why would we check for a `name != ""` if it has been launched by the engine? Wouldn't we instead just error even if name is set or empty?

pkg/adaptation/setup.go
joe0BAB: Is there any down side of never calling close on this? It would just remain open until the IPC stuff is cleaned up? (the server handler in there is the last one referencing it)

I guess it's just fine to remove that line? Agree with your concerns.

pkg/adaptation/setup.go
joe0BAB: Let me change the error messages. I believe they are misleading and might be the cause for your confusion.

pkg/adaptation/setup.go
Benehiko: the garbage collector will remove it once there isn't anyone using it anymore. But at least then it won't cause a potential panic.

pkg/adaptation/setup.go
joe0BAB: Actually you're right, the first check is wrong and things need to change a bit.

Benehiko (approved):

PR #45, opened 2025-06-25
feat(chore): pattern validation

joe0BAB:
The concept of a pattern that can be used to match a (component) ID should align with the existing logic around IDs.

That includes:
- validation (in alignment with the existing pattern match and ID validation logic)
- re-using the match function for checking if an ID matches a pattern


pkg/secrets/pattern_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParsePattern(t *testing.T) {
+	tests := []struct {
+		name   string
+		input  string
+		outErr bool
```
Benehiko:
```suggestion
		mustError bool
```

maybe we should be more explicit? took me a while to get why we are saying "false" to something that looked valid :laughing:

Benehiko (approved):

pkg/secrets/pattern_test.go
joe0BAB: fair, I also tripped over this, will change it

PR #46, opened 2025-06-25
store/keychain: support linux

### LINKED ISSUE #5: Support OS keychain [macos, linux, windows]

Benehiko:
This patch adds Linux support by integrating with the freedesktop secret service (org.freedesktop.secrets) API over DBUS.

It requires that the host have `gnome-keyring` and `dbus-x11` installed and running. Alternatives such as `kwallet` should also work.

Relates to https://github.com/docker/secrets-engine/pull/16
Related #54


store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,284 @@
...
+	"fmt"
+	"slices"
+
+	"github.com/docker/secrets-engine/store"
+	"github.com/keybase/dbus"
+	kc "github.com/keybase/go-keychain/secretservice"
+)
+
+const (
+	// the default collection in most X11 sessions would be 'login'
```
joe0BAB, line 18: Maybe add some docs reference? We would we need to look to validate that object path?

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,284 @@
...
+	loginKeychainObjectPath = dbus.ObjectPath("/org/freedesktop/secrets/collection/login")
+)
+
+func (k *keychainStore[T]) itemAttributes(id store.ID) map[string]string {
```
joe0BAB: Same as last time: Can we turn this into a `newAttributes` standalone function?

I don't see why it would need to be linked to `keychainStore`. It just creates a new map.

joe0BAB: (or maybe `newAttributesMap`, I'll leave the `new<something>` naming up to you)

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,284 @@
...
+	return attributes
+}
+
+// getDefaultCollection gets the secret service collection dbus object path.
+//
+// It prefers the loginKeychainObjectPath, since most users on X11 would have
+// this available.
+//
+// As a fallback it queries the secret service for the default collection and
+// returns that instead.
```
joe0BAB: I believe a reference where that `"org.freedesktop.Secret.Service.Collections"` is coming from wouldn't hurt.

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,284 @@
...
+func (k *keychainStore[T]) getDefaultCollection(service *kc.SecretService) (dbus.ObjectPath, error) {
```
joe0BAB: I don't see where we need that reference to `keychainStore`. Maybe turn it into a standalone helper function? (ie just remove that `keychainStore` ref I mean)

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,284 @@
...
+	return defaultKeychainObjectPath, nil
+}
+
+var errCollectionLocked = errors.New("collection is locked")
+
+// isCollectionLocked verifies if the collection is locked.
+//
+// It returns the errCollectionLocked error by default if the collection is locked.
+// On any other error, it returns the underlying error instead.
+func (k *keychainStore[T]) isCollectionLocked(service *kc.SecretService) error {
```
joe0BAB: I don't see where we need that reference to keychainStore. Maybe turn it into a standalone helper function? (ie just remove that keychainStore ref I mean)

joe0BAB: Also as this is the second time that pattern appears, maybe that calls for a restructuring? Eg we could introduce a struct like this

```
type dbusHelper struct {
   service *kc.SecretService
}
```
and then use that to bundle all our dbus helpers. Maybe that would be more appropriate?

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,284 @@
+package keychain
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+
+	"github.com/docker/secrets-engine/store"
+	"github.com/keybase/dbus"
```
joe0BAB, line 13: Meh, it's a bit disappointing that this is already in "public archive" state. I guess you already evaluate why we shouldn't use https://github.com/godbus/dbus instead? Would the price for that copying/re-implementing https://github.com/keybase/go-keychain/blob/master/secretservice/secretservice.go ourselves?

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,292 @@
...
+
+	return defaultKeychainObjectPath, nil
+}
+
+var errCollectionLocked = errors.New("collection is locked")
+
+// isCollectionLocked verifies if the collection is locked.
+//
+// It returns the errCollectionLocked error by default if the collection is locked.
+// On any other error, it returns the underlying error instead.
```
joe0BAB, line 100: it's a bit odd doing this vs just returning `bool, error`. I'd even argue if you'd return `bool, error`, you could drop this documentation and everything would be more self explanatory. I believe regarding how much code returning `error` vs `bool, error` it doesn't really change much, I checked and it seems roughly equivalent in case that was your motivation

store/keychain/keychain_linux.go
Benehiko: I did look into it a bit, but didn't see what the major drawback code-wise would be - time would be the problem here.

store/keychain/keychain_linux.go
Benehiko: I stuck with the `secrets.ID.IsValid()` pattern, i don't mind which one we go with

store/keychain/keychain_linux.go
Benehiko: i'd like to first get a base line working with CI + tests before I do more refactors

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,284 @@
...
+	items, err := service.SearchCollection(objectPath, attributes)
+	if err != nil {
+		return err
+	}
+
+	if len(items) == 0 {
+		return nil
+	}
+
+	return service.DeleteItem(items[0])
```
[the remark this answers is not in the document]
joe0BAB: Thanks chatgpt, I was about to argue something similar 😂

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,292 @@
...
+	service, err := kc.NewService()
+	if err != nil {
+		return err
+	}
+
+	session, err := service.OpenSession(kc.AuthenticationDHAES)
+	if err != nil {
+		return err
+	}
+	defer service.CloseSession(session)
```
joe0BAB, line 126: I see this pattern repeating at the top of every public/exported function. Might be a hint for restructuring / feels like a smell that something isn't structured properly.

store/keychain/keychain_linux.go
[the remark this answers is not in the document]
Benehiko: won't do - want to avoid major refactors right now. It is also possible we would replace this in future anyway.

store/keychain/keychain_linux.go
Benehiko: already discussed in call - will do a follow-up PR documenting decisions

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,314 @@
...
+	items, err := service.SearchCollection(objectPath, attributes)
+	if err != nil {
+		return err
+	}
+
+	if len(items) == 0 {
+		return nil
+	}
+
+	return service.DeleteItem(items[0])
```
joe0BAB, line 153: Would it be an error also if `len(items) != 1`? Would that mean someone else tempered with our thing?

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,314 @@
...
+//
+// It sets the `service:group` and `service:name` attributes as well as the
+// secret id.
+func newItemAttributes[T store.Secret](id store.ID, k *keychainStore[T]) map[string]string {
+	attributes := map[string]string{
+		"service:group": k.serviceGroup,
+		"service:name":  k.serviceName,
+	}
+	if id.String() != "" {
+		attributes["id"] = id.String()
```
joe0BAB: What are the implications of not setting this or setting it to empty string? Would that be an error we should check? 

I'm wondering mostly why we need to check `id.String() != ""` in the first place. Shouldn't that in reality always be true?

It feels a bit we either should move that check up the chain and check as early as possible (maybe we could do similar as for `secret.ID` that `secret.ParseID(id string) (ID, error)` guarantees secret IDs by design always are valid or also return an error here if it's empty. But only having that check as is smells a bit.

store/keychain/keychain_linux.go
```diff
@@ -0,0 +1,314 @@
...
+	items, err := service.SearchCollection(objectPath, attributes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to search collection: %w", err)
+	}
+
+	if len(items) == 0 {
+		return nil, store.ErrCredentialNotFound
+	}
+
+	value, err := service.GetSecret(items[0], *session)
```
joe0BAB, line 197: Same as above, should we check for length 1?

store/keychain/keychain_linux.go
Benehiko: attributes are used to do searches. It's not too obvious here, but basically you can do a search by service:group and service:name only - e.g. list all secrets for our app/group.

But adding an empty "id" will give you nothing, since then it means you stored an item with id = "".

Benehiko: in reality we could accept id as a string and not `store.ID` here.

store/keychain/keychain_linux.go
Benehiko: we can be more explicit i guess

store/keychain/keychain_linux.go
joe0BAB: But calling eg `Get(ctx, "")` doesn't make any sense in the first place, so maybe we could just move the check up there? (and then return an error "empty secret id" or something)

store/keychain/keychain_linux.go
joe0BAB: It could be ok, but code like this always makes me wonder, what if there are more items?

It's not blocking to me, just a hiccup when reading.

Benehiko: I'm not sure about this now. Let's say we find more than 1 item for some odd reason. How can the application recover from that if we error here?

Benehiko: since the ID is so exact, i wonder how you can even store two items alongside one another since the default behavior is _replace_.

store/keychain/keychain_linux.go
Benehiko: It's not `Get("")`, it's `GetAll` that is calling this

Benehiko: no, we can't error here.

joe0BAB (approved):

PR #48, opened 2025-06-26
Fix registration logic in plugin runtime

joe0BAB:
Returning an error was missing.


pkg/adaptation/registration_test.go
```diff
@@ -0,0 +1,192 @@
...
+	tests := []struct {
+		name string
+		v    pluginCfgInValidator
+		test func(t *testing.T, r *registrationLogic, chResult chan registrationResult)
+	}{
+		{
+			name: "can only register once",
+			v:    mockValidatorOK(t),
+			test: func(t *testing.T, r *registrationLogic, chResult chan registrationResult) {
+				out, err := r.register(t.Context(), mockPluginCfgIn)
```
Benehiko, line 69: Woah, testing.T has context now? TIL https://pkg.go.dev/testing#T.Context
added in Go 1.24 :exploding_head:

Benehiko (approved):

pkg/adaptation/registration_test.go
joe0BAB: Makes me feel much safer in live knowing that this could be used to teardown tests more gracefully :D

PR #59, opened 2025-06-27
feat(sdk): plugin runtime to handle externally launched plugins

### LINKED ISSUE #71: Reliably shutdown IPC (plugin system)
Shutting down IPC is non-trivial as we're using a multiplexed socket. As extra difficulty from this, we cannot `server.Close()`/`server.Shutdown()` directly as one normally would, as that closes the underlaying connection entirely as well. Instead, we have to make sure all pending calls are given a bit of time to finish and then step by step close all layers.

The solution right now relies on `yamux.Session.NumStreams()`. There are still open questions how reliable that is or if there is a better way (eg by wrapping the low level listener/socket and draining server connections in a better way).

joe0BAB:
Implements plugin runtime for externally launched plugins. Also improves IPC shutdown behavior to more gracefully wait for pending calls to finish before shutting everything down (also see #71 for future work on this)

e78a764e9b2089a7a735996eb8fab330fc4abc04 closes #71


pkg/adaptation/plugin.go
```diff
@@ -24,3 +32,59 @@ func getPluginRegistrationTimeout() time.Duration {
 	defer timeoutCfgLock.RUnlock()
 	return pluginRegistrationTimeout
 }
+
+var (
+	_ = (secrets.Resolver)((*plugin)(nil))
```
Benehiko: avoid these kind of type assertions, they are less readable than
```go
var _ secrets.Resolver = &plugin
```

The compiler will automatically discard the allocation since we are assigning it to `_`. In Go `_` means discard, whereas in JavaScript `_` would be the variable name :rofl:

pkg/adaptation/plugin_test.go
```diff
@@ -0,0 +1,196 @@
...
+				assert.NoError(t, s.Run(context.Background()))
+				<-doneRuntime
+			},
+		},
+		{
+			name: "cancelling plugin.run() shuts down the runtime",
+			test: func(t *testing.T, l net.Listener, conn net.Conn) {
+				doneRuntime := make(chan struct{})
+				donePlugin := make(chan struct{})
+				downRuntime := make(chan struct{})
```
Benehiko: these look a bit funky to me. I don't want to nit too much on this, but is there a way to clean up how we are doing this logic? perhaps we can replace with a singular `sync.WaitGroup{}` instead?

pkg/adaptation/plugin_test.go
```diff
@@ -0,0 +1,196 @@
...
+					_, err := p.GetSecret(t.Context(), secrets.Request{ID: mockSecretID})
+					assert.ErrorContains(t, err, "id mismatch")
+					assert.NoError(t, p.close())
+					close(doneRuntime)
+				}()
+
+				s, err := p.New(newMockedPlugin(WithID("rewrite-id")), p.WithPluginName("my-plugin"), p.WithConnection(conn))
+				require.NoError(t, err)
+				assert.NoError(t, s.Run(context.Background()))
+				<-doneRuntime
```
Benehiko: waiting on a channel like this also makes me paranoid. If something goes wrong the tests could wait for 10minutes until the test gets killed.

Would be safer to wrap these in a select with a timeout.

pkg/adaptation/plugin_test.go
joe0BAB: rewrote the tests and linearized them (plus added timeouts)

pkg/adaptation/plugin_test.go
joe0BAB: rewrote the tests and linearized them (plus added timeouts)

internal/ipc/design.md
```diff
@@ -51,5 +51,30 @@ Yamux is a full-featured, multiplexing protocol that allows multiple streams to
 Using Yamux we get Go's `net/http` out-of-the-box.
 
 
+## Decisions
```
joe0BAB, line 54: I'll move that section around in a follow up PR (didn't want to expand this one too much)

internal/ipc/design.md
```diff
@@ -51,5 +51,30 @@ Yamux is a full-featured, multiplexing protocol that allows multiple streams to
...
+Also connect rpc is part of CNCF ([source](https://www.cncf.io/projects/connect-rpc/)).
+
+Potential drawbacks: Performance
+
+Using [nri/net/multiplex](https://github.com/containerd/nri/tree/main/pkg/net/multiplex) with [ttrpc](https://github.com/containerd/ttrpc) probably would be the most performance optimised solution. 
+It re-uses one stream over the multiplexed socket per direction and has overhead of the http protocol as protobuf gets streamed directly over the multiplexer. 
+However, it has stopped evolving and e.g. has not caught up on latest improvements on protobuf. 
+Another major downside is that it's mainly Go only, i.e., supporting alternative languages for plugins would come at a high cost.
+We argue that in our use case, as the networking only happens locally so the overhead of grpc over http and the cost of opening a new yamux stream are negligible.
+In addition, the main performance bottleneck will be within the actual plugins, e.g., because authentication needs to happen, disk access or because (slow) external networking access is required.
```
Benehiko:
```suggestion
Using [nri/net/multiplex](https://github.com/containerd/nri/tree/main/pkg/net/multiplex) with [ttrpc](https://github.com/containerd/ttrpc) probably would be the most performant solution. 
It re-uses one stream over the multiplexed socket per direction and does not have the overhead of the HTTP protocol as Protobuf gets streamed directly over the multiplexer. 

Although lightweight, it has stopped evolving and has not caught up to the latest improvements on Protobuf. 
Another major downside is that it's mainly Go only. Plugins written in a different language would come at a high cost.

We argue that in our use case since the networking only happens locally the overhead of GRPC over HTTP and the cost of opening a new yamux stream are negligible. In addition, the main performance bottleneck will be within the actual plugins due to IO operations, additional upstream network requests and potentially authentication.
```

internal/ipc/ipc.go
```diff
@@ -120,6 +147,7 @@ func createYamuxedClient(session *yamux.Session) *http.Client {
 		DialContext: func(context.Context, string, string) (net.Conn, error) {
 			return session.Open()
 		},
+		DisableKeepAlives: true,
```
Benehiko, line 154: Just add a comment here to say why this is disabled.

pkg/adaptation/plugin.go
```diff
@@ -24,3 +32,59 @@ func getPluginRegistrationTimeout() time.Duration {
...
+	sync.Mutex
+	base           string
+	pattern        secrets.Pattern
+	version        string
+	pluginClient   resolverv1connect.PluginServiceClient
+	resolverClient resolverv1connect.ResolverServiceClient
+	close          func() error
+}
+
+// newExternalPlugin Create a plugin (stub) for an accepted external plugin connection.
```
Benehiko:
```suggestion
// newExternalPlugin creates a plugin (stub) for an accepted external plugin connection.
```

pkg/adaptation/plugin_test.go
```diff
@@ -0,0 +1,234 @@
...
+		{
+			name: "plugin returns error on GetSecret",
+			test: func(t *testing.T, l net.Listener, conn net.Conn) {
+				m := &mockExternalRuntime{l: l, done: make(chan struct{})}
+				go m.run()
+
+				s, err := p.New(newMockedPlugin(WithID("rewrite-id")), p.WithPluginName("my-plugin"), p.WithConnection(conn))
+				require.NoError(t, err)
+				runErr, cancel := runAsyncWithTimeout(t.Context(), s.Run)
+				defer cancel()
```
Benehiko:
```suggestion
				t.Cleanup(cancel)
```

pkg/adaptation/plugin_test.go
```diff
@@ -0,0 +1,234 @@
...
+		{
+			name: "create external plugin",
+			test: func(t *testing.T, l net.Listener, conn net.Conn) {
+				m := &mockExternalRuntime{l: l, done: make(chan struct{})}
+				go m.run()
+
+				s, err := p.New(newMockedPlugin(), p.WithPluginName("my-plugin"), p.WithConnection(conn))
+				require.NoError(t, err)
+				runErr, cancel := runAsyncWithTimeout(t.Context(), s.Run)
+				defer cancel()
```
Benehiko:
```suggestion
				t.Cleanup(cancel)
```

pkg/adaptation/plugin_test.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func Test_newExternalPlugin(t *testing.T) {
+	tests := []struct {
+		name string
+		test func(t *testing.T, listener net.Listener, conn net.Conn)
+	}{
+		{
+			name: "create external plugin",
+			test: func(t *testing.T, l net.Listener, conn net.Conn) {
```
Benehiko, line 82:
```suggestion
			test: func(t *testing.T, l net.Listener, conn net.Conn) {
			    t.Helper()
```

might need to add `t.Helper()` for better test failure logs. It will attribute this function to the test run instead of separately.

Benehiko (approved):

PR #60, opened 2025-06-27
fix(sdk): simplify interface and correctly handle shutdown scenarios

joe0BAB:
Another iteration on getting the IPC reliable and stable. This round: Correctly handle when either side terminates unexpectedly.


internal/ipc/ipc.go
```diff
@@ -4,36 +4,29 @@ import (
...
 	session, err := yamux.Client(sockConn, nil)
 	if err != nil {
 		return nil, nil, fmt.Errorf("creating yamux client: %w", err)
 	}
-	i, c := newMuxedIPC(session, handler)
+	i, c := newMuxedIPC(session, handler, onServerClosed)
 	return i, c, nil
 }
 
-type IPC interface {
```
joe0BAB, line 23: Getting rid of this and just boiling things down to a callback+`io.Closer`+`*http.Client` I believe is a major win and makes me feel much more confident about the implementation (more boring, less fancy).

joe0BAB:
On my machine running this with `go test -run Test_newExternalPlugin -count 10000 -failfast` didn't show anything. I think these tests should be reliable.

pkg/adaptation/setup.go
```diff
@@ -36,16 +38,16 @@ func Setup(conn net.Conn, v setupValidator) (*SetupResult, error) {
 	})
 	registrator := newRegistrationLogic(v, chRegistrationResult)
 	httpMux.Handle(resolverv1connect.NewEngineServiceHandler(&RegisterService{registrator}))
-	i, c, err := ipc.NewRuntimeIPC(conn, httpMux)
+	chIpcErr := make(chan error, 1)
+	i, c, err := ipc.NewRuntimeIPC(conn, httpMux, func(err error) {
+		if errors.Is(err, io.EOF) {
+			logrus.Infof("Connection to plugin %v closed", v.name)
+		}
+		chIpcErr <- err
```
Benehiko, line 46: would it make sense to set `err = nil` here too on `io.EOF`?

joe0BAB: not sure yet, it depends on the code in the follow up PR that will use this. Let's leave as is for now, I'll need to work on this in follow up PRs anyway still.

Benehiko (approved):

PR #65, opened 2025-06-27
store/keychain(docs): readme, ci setup and tests

### LINKED ISSUE #5: Support OS keychain [macos, linux, windows]

Benehiko:
- Document the keychain CI setup and tests.
- Rename the `design` document.
- Move keychain docs to the `store/docs` directory
- Add a `store/test.md` file
- Add top level `docs/ci/keychain.md` file


docs/ci/keychain.md
```diff
@@ -0,0 +1,67 @@
...
+This will use `buildkit` to target only the `ubuntu-24-gnome-keyring` label inside
+the `store/Dockerfile`.
+
+Each backend has a script to start them up and ensure they are running before
+any Go tests even run. They are located in `store/scripts/gnome-keyring` and
+`store/scripts/kdewallet`.
+
+### Fedora
+
+On Fedora we install `gnome-keyring`, `kf6-kwallet` and `dbus-daemon`.
```
joe0BAB: remove: this doesn't bring extra value over that equivalent line in the Dockerfile

docs/ci/keychain.md
```diff
@@ -0,0 +1,67 @@
...
+We require `dbus-daemon` since it was removed in favor of `dbus-broker` over
+`systemd`. We then use `dbus-daemon` to start the `dbus` service and get the
+connection address for `gnome-keyring` and `kwalletd6`.
+
+Fedora 43 is the latest at the time of writing and has the latest packages and
+changes to `kdewallet` and `dbus`.
+
+### Ubuntu
+
+On Ubuntu we install `libglib2.0-bin`, `dbus`, `gnome-keyring`, `kwalletmanager`.
```
joe0BAB: remove: this doesn't bring extra value over that equivalent line in the Dockerfile

docs/ci/keychain.md
```diff
@@ -0,0 +1,67 @@
+# Keychain CI
+
+The keychain tests are split between macOS, Linux and Windows. There are two
+make commands: `make keychain-unit-tests` and `make keychain-linux-unit-tests`.
+
+For local development, it would make the most sense to just run `keychain-unit-tests`
```
joe0BAB: Move up into sub module repo and add a section "local development"

docs/ci/keychain.md
```diff
@@ -0,0 +1,67 @@
...
+be a headache to setup and maintain and we don't have access to a variety of
+distros.
+
+Linux has two popular keychain backends: `gnome-keyring-daemon` and `kdewallet`.
+To cover a variety of environments, we setup `Ubuntu 24.04` and `Fedora 43` with
+both backends in different test runs.
+
+To test this locally you can run `DOCKER_TARGET=ubuntu-24-gnome-keyring make keychain-linux-unit-tests`.
+This will use `buildkit` to target only the `ubuntu-24-gnome-keyring` label inside
+the `store/Dockerfile`.
```
joe0BAB: move to section "local development" / "local testing"

docs/ci/keychain.md
```diff
@@ -0,0 +1,67 @@
...
+
+Each backend has a script to start them up and ensure they are running before
+any Go tests even run. They are located in `store/scripts/gnome-keyring` and
+`store/scripts/kdewallet`.
```
joe0BAB: Could we instead have a mermaid diagram that depicts the whole setup?

In my head this looks something like this:

image

docs/ci/keychain.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Fedora
+
+On Fedora we install `gnome-keyring`, `kf6-kwallet` and `dbus-daemon`.
+We require `dbus-daemon` since it was removed in favor of `dbus-broker` over
+`systemd`. We then use `dbus-daemon` to start the `dbus` service and get the
+connection address for `gnome-keyring` and `kwalletd6`.
+
+Fedora 43 is the latest at the time of writing and has the latest packages and
+changes to `kdewallet` and `dbus`.
```
joe0BAB: remove (that's not gonna age well)

docs/ci/keychain.md
```diff
@@ -0,0 +1,67 @@
...
+
+On a high-level, each script does:
+
+1. Check relevant binaries available
+2. Creates the necessary files for the keyring daemon
+3. Starts the dbus daemon
+4. Sets the `DBUS_SESSION_BUS_ADDRESS` environment variable
+5. Starts the keyring
+6. Waits for the keyring to be active by polling dbus over `gdbus`
+7. Check if the registered `org.freedesktop.secrets` backend matches what we expect.
```
joe0BAB: Move at the top of the script? (how when reading the script would I know that the docs of it are here?)

docs/ci/keychain.md
```diff
@@ -0,0 +1,67 @@
...
+- `ubuntu-24-kdewallet`
+- `fedora-43-gnome-keyring`
+- `fedora-43-kdewallet`
+
+Each test uses `buildkit` to setup a base image layer specified in `store/Dockerfile`.
+There are two base image layers `ubuntu-24.04` and `fedora-43`. They install
+all the necessary packages which gives us caching.
+
+Then for each of the four tests listed above we invoke the relevant scripts stored
+in `store/scripts`.
```
joe0BAB: I'd suggest to merge that with the diagram suggested above (getting readers to understand a process is a very good candidate for diagrams!)

store/keychain/design.md
```diff
@@ -0,0 +1,57 @@
+# Keyring
+
+The keyring package supports macOS, Linux and Windows applications
+store, delete and retrieve secrets in a secure way.
```
joe0BAB:
```suggestion
.
```
(obvious?)

store/keychain/design.md
```diff
@@ -0,0 +1,57 @@
...
+
+The goals are:
+
+- Implement the `store.Store` interface to be tightly coupled with the secrets engine
+- Support multiple platforms namely: Linux, macOS, Windows
+- Should be a standalone library
+- Support credentials of any data structure
```
joe0BAB: remove (we already know all that from above / top-level readme/go.mod)

store/keychain/design.md
```diff
@@ -0,0 +1,57 @@
...
+
+The `keyring` package is the successor of the [docker-credential-helpers](https://github.com/docker/docker-credential-helpers/).
+
+It solves a lot of the drawbacks of its predecessor, such as:
+
+- Generic credentials
+- Broader use cases (not just a registry credential store)
+- Native Go library
+- Bundles with your application binary
+- More secure
```
joe0BAB: if you want to write that, you have to elaborate on that (otherwise to many people that could be read just as an accusation)

Eg you could do like this: `More secure: reason 1, reason 2...`

joe0BAB: I'd suggest to reframe this a bit and instead of explicitly say "successor", have a section "secrets-engine/store/keychain vs docker-credential-helpers" and turn it more into a comparison. Feels more sensitive (I mean it's obvious that we created ours because we didn't think the cred helper was a good idea, so I'd not rub it too much into the readers face but just stick to the facts/comparison).

store/keychain/design.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Linux
+
+Users running Linux with a Desktop Environment usually have access to the
```
joe0BAB:
```suggestion
Users running Linux with a desktop environment usually have access to the
```
?

store/keychain/design.md
```diff
@@ -0,0 +1,57 @@
...
+[`org.freedesktop.secrets`](https://specifications.freedesktop.org/secret-service-spec/latest/index.html) API
+via `gnome-keyring` or `kdewallet`.
+
+Usually the `pam_gnome_keyring.so` and `pam_kwallet5.so` would hook into PAM
+and automatically unlock the 'login' keyring once the user does a login to their system.
+If the 'login' keyring does not exist, it will be created using the user's login password.
+If the 'login' keyring is the first keyring created, it will be set as the default.
+For more information regarding `gnome-keyring` and PAM, please refer to the
+[GnomeKeyring documentation](https://wiki.gnome.org/Projects/GnomeKeyring/Pam)
```
joe0BAB: Very useful!

store/keychain/design.md
```diff
@@ -0,0 +1,57 @@
...
+- Secure cryptographic communication over the `dbus` connection
+- Easy to use API
+
+Some of the drawbacks are:
+
+- Relies on a forked archived [keybase/dbus](https://github.com/keybase/dbus) library
+- Low contribution activity
+
+In future we might update the [keybase/dbus](https://github.com/keybase/dbus) with
+a more up to date and maintained version, such as [godbus/dbus](https://github.com/godbus/dbus).
```
joe0BAB: This section effectively is a decision log. I'm wondering if we could make that more clear (eg adding some sub-header "<date> deciding on key-base/go-keychain") or move it into something separate (so docs that should always be up-to-date and decision which captures a moment in time don't mix too much).

store/keychain/design.md
Benehiko: i mean, in Docker we are succeeding the credential helper with the keychain library as part of solving auth. I wanted to make that clear here somehow.

store/keychain/design.md
Benehiko: okay, I've added a decision log document

docs/ci/keychain.md
```diff
@@ -0,0 +1,75 @@
...
+variety of distros.
+
+Linux has two popular keychain backends: `gnome-keyring-daemon` and `kdewallet`.
+To cover a variety of environments, we run `Ubuntu` and `Fedora` tests using
+Docker with different keychain backend.
+
+Each keychain backend has a script to start them up and ensure they are running
+before any Go tests even run. They are located in `store/scripts/gnome-keyring`
+and `store/scripts/kdewallet`.
+
```
joe0BAB, line 37:
```suggestion
## linux vm / buildkit

For linux, instead of dealing with different GH linux distro runners, we use docker distro images and buildkit. Thus instead of having to manage different GH linux distros and customizing them, we only have to maintain a multistage Dockerfile which is way simpler.
```

I'd add a sub-header and elaborate a bit on the benefits of the buildkit setup for linux (eg as above but feel free to improve).

docs/ci/keychain.md
```diff
@@ -0,0 +1,75 @@
...
+    D -->|fedora| E(Run Test)
+    D -->|ubuntu| F(Run Test)
+```
+
+The keychain tests are split between macOS, Linux and Windows. There are two
+make commands: `make keychain-unit-tests` and `make keychain-linux-unit-tests`.
+
+Since there are so many different scenarios in Linux, the GH runners would
+be a headache to setup and maintain and we don't have access to the necessary
+variety of distros.
```
joe0BAB: I'd remove that here in favor of the suggested sub-section header below.

docs/ci/keychain.md
```diff
@@ -0,0 +1,75 @@
...
+
+Linux has two popular keychain backends: `gnome-keyring-daemon` and `kdewallet`.
+To cover a variety of environments, we run `Ubuntu` and `Fedora` tests using
+Docker with different keychain backend.
+
+Each keychain backend has a script to start them up and ensure they are running
+before any Go tests even run. They are located in `store/scripts/gnome-keyring`
+and `store/scripts/kdewallet`.
```
joe0BAB: I'd suggest to also move below the suggested linux sub header.

store/docs/keychain/design.md
```diff
@@ -0,0 +1,37 @@
+# Keyring
+
+The `keyring` package is the successor of the [docker-credential-helpers](https://github.com/docker/docker-credential-helpers/)
+and supports macOS, Linux and Windows applications.
```
joe0BAB: I'd frame this more neutral, not use the word successor (for us this is true, but maybe for others it isn't) and instead phrase this more as a comparison to docker-credential-helpers.

store/docs/test.md
```diff
@@ -0,0 +1,36 @@
+# Test
+
+The store can be tested using the `go test` command. For the `keychain` package
+this might require further tweaks depending on the OS you are running on.
+
+## Keychain tests
+
+For local development, it would make the most sense to just run `keychain-unit-tests`
+since it's simply invoking `go test` for only the `keychain` package. CGO is
+enabled to support macOS.
```
joe0BAB, line 10:
```suggestion
For local development, run `make keychain-unit-tests`.
This will run all tests in `keychain` package for your OS.
```

store/docs/test.md
```diff
@@ -0,0 +1,36 @@
...
+
+To test Linux on other OSs (like macOS and Windows) or isolated from the host,
+you can use the `make keychain-linux-unit-tests` command. It uses Docker to
+spin up the necessary environment required to run the test.
+
+```console
+DOCKER_TARGET=ubuntu-24-gnome-keyring make keychain-linux-unit-tests
+```
```
joe0BAB:
```suggestion
The cross distro linux tests can be run via:
` ` `console
DOCKER_TARGET=ubuntu-24-gnome-keyring make keychain-linux-unit-tests
` ` `
```

I'd simplify

docs/ci/keychain.md
```diff
@@ -0,0 +1,75 @@
...
+We require `libglib2.0-bin` since we require `gdbus` CLI to talk to the `dbus`
+APIs.
+
+## Understand the Process
+
+The GitHub action spins up three jobs, `linux-keychain`,`test-macos` and `test-windows`.
+Each job is run in its own runner. For example `test-macos` runs inside a `macos`
+runner. The Linux job is run on an Ubuntu runner, but the actual tests are run
+through Docker Buildkit.
+
```
joe0BAB, line 47: Maybe we could move the linux section from above in here and the have all linux related here?

joe0BAB (approved):
LGTM! Non-blocking: I'd streamline the `ci/keychain.md` a bit, also the `test.md` still could be slimmed a bit.

PR #67, opened 2025-06-30
store/keychain: windows support

### LINKED ISSUE #5: Support OS keychain [macos, linux, windows]

### LINKED ISSUE #51: Windows Keychain
Support Windows keychain using the https://github.com/danieljoos/wincred library. 

Note:
> The size of a credential blob is limited to 2560 Bytes by the Windows API.

### LINKED ISSUE #52: Windows Keychain GHA CI PR Checks

Benehiko:
Closes #51 
Closes #52 

This patch adds support for the [Windows Credential Manager API](https://learn.microsoft.com/en-us/windows/win32/api/wincred/). It uses the [danieljoos/wincred](https://github.com/danieljoos/wincred) library. All credentials stored by the Windows Credential Manager are encoded in UTF16 format - since most applications accessing the credential manager expect this format.

Relates to #16


store/keychain/keychain_windows.go
```diff
@@ -0,0 +1,160 @@
...
+		id, err := store.ParseID(strings.ReplaceAll(cred.TargetName, onlyLabelPrefix, ""))
+		if err != nil {
+			return nil, err
+		}
+
+		gc, err := wincred.GetGenericCredential(cred.TargetName)
+		if err != nil {
+			return nil, mapWindowsCredentialError(err)
+		}
+		if err := secret.Unmarshal(gc.CredentialBlob); err != nil {
```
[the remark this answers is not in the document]
Benehiko: no, blob is not part of list.

store/keychain/keychain_windows.go
```diff
@@ -0,0 +1,191 @@
...
+func (k *keychainStore[T]) Delete(ctx context.Context, id store.ID) error {
+	if err := id.Valid(); err != nil {
+		return err
+	}
+
+	g, err := wincred.GetGenericCredential(k.itemLabel(id))
+	if err != nil && !errors.Is(err, wincred.ErrElementNotFound) {
+		return mapWindowsCredentialError(err)
+	}
+	if g == nil {
```
joe0BAB: For readability: could we change that to `g == nil || errors.Is(err, winced.ErrElementNotFound)`? (it probably doesn't change logic but makes reading the code easier)

joe0BAB: But also, should deleting a secret that doesn't exist return an error? 🤔 

Regardless of the answer, is that consistent across all OS and can we add verifying that it's consistent to the unit tests that check across all OS?

store/keychain/keychain_windows.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil
+	}
+
+	err = g.Delete()
+	if err != nil && !errors.Is(err, wincred.ErrElementNotFound) {
+		return mapWindowsCredentialError(err)
+	}
+	return nil
```
joe0BAB, line 62: The answer above might affect this as well.

store/keychain/keychain_windows.go
Benehiko: It's not consistent, windows can return an error yes :see_no_evil: 
https://learn.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-creddeletew#return-value

joe0BAB: What I meant is, can you add to `keychain_test.go` a test case that verifies deleting a non existing secret behaves the same across all implementations? We don't have that test right now.

store/keychain/keychain_windows.go
```diff
@@ -0,0 +1,191 @@
...
+			return nil, err
+		}
+
+		gc, err := wincred.GetGenericCredential(cred.TargetName)
+		if err != nil {
+			return nil, mapWindowsCredentialError(err)
+		}
+
+		decoder := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewDecoder()
+		blob, _, err := transform.Bytes(decoder, gc.CredentialBlob)
```
joe0BAB, line 140: non blocking: I'd probably move that into a function just for the sake as I wouldn't trust myself to repeat that without an error 😅 (ie main reason would be safety, not dry)

joe0BAB (approved):
LGTM! Could we add the missing test in `keychain_test.go` to the TODO/issue that is going to improve those tests? I believe we already have that somewhere.

store/keychain/keychain_windows.go
Benehiko: We can add more tests, but i don't think it's consistent in terms of when it errors. e.g. on success it returns nil error, but if it does error it can be for various reasons. 

I'll do that in a follow-up when refactoring the tests.

Benehiko: simplified the if check - instead of querying first, I let the delete handle the `ErrNotFound`.

store/keychain/keychain_windows.go
Benehiko: done

PR #69, opened 2025-06-30
fix(sdk): race condition issues on stub resolver implementation

joe0BAB:
There was a potential race condition where server a secret resolve request could race against the setup not having completed yet. Also added a missing check that plugins can't rewrite secret IDs.


plugin/resolver.go
```diff
@@ -16,10 +18,20 @@ import (
...
 
 type resolverService struct {
-	resolver secrets.Resolver
+	resolver            secrets.Resolver
+	setupCompleted      chan struct{}
+	registrationTimeout time.Duration
 }
 
 func (r *resolverService) GetSecret(ctx context.Context, c *connect.Request[resolverv1.GetSecretRequest]) (*connect.Response[resolverv1.GetSecretResponse], error) {
+	logrus.Infof("GetSecret request (ID %q)", c.Msg.GetSecretId())
```
Benehiko: i guess we don't want this log here? (should we mark it as a debug level?)

plugin/resolver_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+func withMockResolverError(err error) mockResolverOption {
+	return func(m *mockResolver) *mockResolver {
+		m.err = err
+		return m
+	}
+}
+
+func (m mockResolver) GetSecret(_ context.Context, request secrets.Request) (secrets.Envelope, error) {
+	assert.Equal(m.t, m.id, request.ID)
```
Benehiko: I'm not sure if doing an assert inside such a function is a good idea. It's also not marked as a `t.Helper` which will make debugging it really difficult.

Benehiko: you can instead just return an error based on an if check, and then assert on the error inside the actual test.

plugin/resolver.go
joe0BAB: let's mark as debug

Benehiko (approved):

PR #70, opened 2025-07-01
feat(sdk): Use server.Shutdown instead of server.Close for graceful shutdown

joe0BAB:
This is step 1/2 fixing the shutdown behavior to a graceful shutdown:

- [x] replace server.Close with server.Shutdown
- [ ] close the socket connection last (follow up PR)


internal/ipc/ipc.go
```diff
@@ -50,18 +68,32 @@ func newIpcServer(l net.Listener, handler http.Handler, onClose func(error) erro
 		if errors.Is(err, http.ErrServerClosed) { // not an error, client closed the connection
 			err = nil
 		}
-		result.err = errors.Join(err, onClose(err))
+		result.err = errors.Join(filterEOF(err), onClose(err)) // EOF: only forward to the onClose handler, but filter out internal forwarding
```
joe0BAB, line 71: This EOF appearing here is new and was caught by the tests. We don't want to leak it into the Close result.

internal/ipc/ipc.go
```diff
@@ -50,18 +68,32 @@ func newIpcServer(l net.Listener, handler http.Handler, onClose func(error) erro
...
 
-func newMuxedIPC(session *yamux.Session, handler http.Handler, onClose func(error)) (*ipcImpl, *http.Client) {
+func newMuxedIPC(session *yamux.Session, handler http.Handler, onClose func(error), option ...Option) (*ipcImpl, *http.Client) {
+	// Note: Calling session.Close() needs to be done as the very last step as it shuts down all IPC!
+
+	cfg := &cfg{shutdownTimeout: defaultShutdownTimeout}
+	for _, o := range option {
+		cfg = o(cfg)
+	}
+	//l := &wrappedListener{Listener: session, done: make(chan struct{})}
```
Benehiko: remove?
```suggestion
	
```

Benehiko (approved):
small nit

PR #76, opened 2025-07-03
feat(sdk): runtime for engine launched plugins

joe0BAB:
This implements the runtime logic for plugins launched by the engine.
I.e., it does all the logic to setup a connection pair, start the plugin process, monitor it and then proceed to the normal setup steps that are equivalent to externally launched plugins.


pkg/adaptation/plugin_windows.go
```diff
@@ -0,0 +1,18 @@
+//go:build windows
+
+package adaptation
+
+import "os/exec"
+
+// TODO
```
joe0BAB: I'll add windows support later

pkg/adaptation/plugin_test.go
```diff
@@ -72,6 +74,63 @@ func (m mockedPlugin) Configure(context.Context, p.RuntimeConfig) error {
...
+	}
+}
+
+func Test_newPlugin(t *testing.T) {
+	tests := []struct {
+		name string
+		test func(t *testing.T)
+	}{
+		{
+			name: "engine launched plugin",
```
joe0BAB, line 93: This is just the happy path test. I'll add tests for the failure cases in follow up PR.

pkg/adaptation/plugin_test.go
```diff
@@ -72,6 +74,63 @@ func (m mockedPlugin) Configure(context.Context, p.RuntimeConfig) error {
...
 
+// TestMain acts as a dispatcher to run as dummy plugin or normal test.
+// Inspired by: https://github.com/golang/go/blob/15d9fe43d648764d41a88c75889c84df5e580930/src/os/exec/exec_test.go#L69-L73
+func TestMain(m *testing.M) {
+	if os.Getenv("RUN_AS_DUMMY_PLUGIN") != "" {
+		DummyPluginProcess()
+	} else {
+		os.Exit(m.Run())
+	}
+}
```
Benehiko: :partying_face:

pkg/adaptation/dummy.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"github.com/docker/secrets-engine/pkg/secrets"
+	"github.com/docker/secrets-engine/plugin"
+)
+
+const (
+	dummyPluginCfgEnv = "DUMMY_PLUGIN_CFG"
+)
+
+func DummyPluginCommand(t *testing.T, cfg dummyPluginCfg) (*exec.Cmd, func() (*dummyPluginResult, error)) {
```
Benehiko: add a godoc for this, also it's exported, do we want that? The tests should be fine if it's unexported right?

pkg/adaptation/dummy.go
```diff
@@ -0,0 +1,154 @@
...
+
+func newDummyPluginCfg(in string) (*dummyPluginCfg, error) {
+	var result dummyPluginCfg
+	if err := json.Unmarshal([]byte(in), &result); err != nil {
+		return nil, fmt.Errorf("failed to decode dummy plugin cfg: %w", err)
+	}
+	return &result, nil
+}
+
+func DummyPluginProcess() {
```
Benehiko: add missing godoc and unexport (?)

pkg/adaptation/dummy.go
```diff
@@ -0,0 +1,154 @@
...
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer cancel()
+	var logBuf bytes.Buffer
+	logrus.SetOutput(&logBuf)
+	cfgStr := os.Getenv(dummyPluginCfgEnv)
+	cfg, err := newDummyPluginCfg(cfgStr)
+	if err != nil {
+		panic(err)
```
Benehiko: perhaps we could change it to use `log.Fatalf` instead of `panic`. Usually panic doesn't give much information.

pkg/adaptation/exec_unix.go
```diff
@@ -0,0 +1,39 @@
+//go:build !windows
```
Benehiko, line 1: i think this might be unnecessary if you have already named the file for `_unix` only.

pkg/adaptation/exec_windows.go
```diff
@@ -0,0 +1,18 @@
+//go:build windows
```
Benehiko, line 1: same here, you have specified the file name `_windows`.

pkg/adaptation/plugin_test.go
```diff
@@ -72,6 +74,63 @@ func (m mockedPlugin) Configure(context.Context, p.RuntimeConfig) error {
...
+
+func Test_newPlugin(t *testing.T) {
+	tests := []struct {
+		name string
+		test func(t *testing.T)
+	}{
+		{
+			name: "engine launched plugin",
+			test: func(t *testing.T) {
```
Benehiko, line 94: add `t.Helper()`

pkg/adaptation/plugin_test.go
```diff
@@ -72,6 +74,63 @@ func (m mockedPlugin) Configure(context.Context, p.RuntimeConfig) error {
...
+				r, err := parseOutput()
+				require.NoError(t, err)
+				require.Equal(t, 1, len(r.GetSecret))
+				assert.Equal(t, mockSecretID, r.GetSecret[0].ID)
+				assert.Equal(t, 1, r.ConfigRequests)
+				require.Equal(t, 1, len(r.Configure))
+				assert.Equal(t, mockEngineName, r.Configure[0].Engine)
+				assert.Equal(t, mockEngineVersion, r.Configure[0].Version)
+
+				r.prettyPrintLogs()
```
Benehiko: this seems like it would potentially make debugging tests more difficult. I think we should rather opt for printing using `t.Logf`. One could set the logger for the dummy plugin to the test logger - then at least you aren't _always_ printing it, only in circumstances where we pass `go test -v`.

pkg/adaptation/exec_unix.go
joe0BAB: the linter complaints otherwise about duplicated function definitions

pkg/adaptation/dummy.go
joe0BAB: Hm, I didn't want to use log/logrus, so it will be easier to distinguish between logs coming from the plugin code vs from the dummy plugin. But we can add a field to the result struct and then even (try) handling things more explicitly (note that there's still that last error from json marshaling to which we can't do much about)

pkg/adaptation/plugin_test.go
joe0BAB: btw, we should actually not do that, adding this actually makes the output worse

I think we've misunderstood the usage of `t.Helper()` a bit

pkg/adaptation/exec_unix.go
Benehiko: ah right, `_unix` is not a valid filename suffix. only `_darwin` | `_linux` is.

pkg/adaptation/plugin_test.go
Benehiko: yeah like we discussed, we aren't executing a sub-function here (as I initially thought) which means `t.Helper` is not what we want.

Benehiko (approved):

PR #78, opened 2025-07-03
feat(sdk): remove configure method from plugin interface

joe0BAB:
We removing the `config` string field and the corresponding `Configure` plugin interface API call (see comments on code for further details).


plugin/stub.go
```diff
@@ -22,17 +22,9 @@ type Plugin interface {
 
 	Config() Config
 
-	Configure(context.Context, RuntimeConfig) error
```
joe0BAB, line 25: The main value initially of this was the `Config` string field of `RuntimeConfig` which we agreed to remove.

There still could be value in passing the `Engine` and `Version` (engine version) values to a plugin. However, as of now we don't have any use case for it plus the mechanics of `Configure` right now are buggy (the runtime never gets informed if a `Configure` call in the stub would fail but the plugin would just shutdown).

So I'm removing it now as it feels like pre-mature optimization. We can always add it back later. But right now it doesn't bring any value but only adds bugs.

pkg/api/resolver/v1/api.proto
```diff
@@ -19,14 +19,12 @@ message RegisterPluginRequest {
 }
 
 message RegisterPluginResponse {
-  // Any plugin-specific data, if present among the engine configuration.
-  string config = 1;
```
joe0BAB, line 23: Removing this field is the main change of the PR, the rest mainly is propagating it through.

Benehiko (approved):

PR #83, opened 2025-07-04
Improved lint and format on multi-platform targets

Benehiko:
This patch fixes linting Go files that rely on CGO. The underlying github.com/docker/secrets-engine/store module offers a keychain package which uses CGO when compiling on darwin. Due to this, the golangci-lint command tries to lint the source code based on the darwin build target which then fails on Linux hosts.

We already support multi-platform builds through BuildKit, unfortunately we need to use CGO as well which won't work.  To support cross-platform linting with CGO we use osxcross.

The workaround is to set up independent layers with platform dependent logic as a "base" layer. This layer contains the Go binary and in the case of darwin the necessary tools to do cross-compilation.

We then need to setup the toolchain so that golangci-lint points to osxcross.


.golangci.yml
```diff
@@ -15,26 +12,56 @@ linters:
...
+      - comments
+      - common-false-positives
+      - legacy
+      - std-error-handling
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
+      - vendor$
+formatters:
```
joe0BAB, line 47: cool!

joe0BAB (approved):

PR #85, opened 2025-07-07
feat(sdk): on shutdown run all close calls in parallel

joe0BAB:
I'm building up the engine logic and this is some tiny functionality that run fast shutdown.


pkg/adaptation/engine.go
```diff
@@ -0,0 +1,20 @@
...
+func parallelStop(plugins []runtime) error {
+	errCh := make(chan error, len(plugins))
+	for _, p := range plugins {
+		go func(pl runtime) {
+			errCh <- pl.Close()
+		}(p)
+	}
+	var errs []error
+	for i := 0; i < len(plugins); i++ {
+		if err := <-errCh; err != nil {
```
Benehiko: so this is blocking, there are no guarantees that `pl.Close()` will return instantly, making this still O(n) since it waits for each `Close()` to write to the channel.

pkg/adaptation/engine.go
```diff
@@ -0,0 +1,20 @@
...
+			errs = append(errs, err)
+		}
+	}
+	return errors.Join(errs...)
+}
```
Benehiko: I think we can simplify this a bit more by using a waitgroup.

```suggestion
func parallelStop(plugins []runtime) error {
	errCh := make(chan error, len(plugins))
	wg := &sync.WaitGroup{}
	for _, p := range plugins {
	        wg.Add(1)
		go func(pl runtime) {
		        defer wg.Done()
			errCh <- pl.Close()
		}(p)
	}
	wg.Wait()
	close(errCh)
	var errs error
	for err := range errCh {
		errs = errors.Join(errs, err)
	}
	return errs
}
```

pkg/adaptation/engine.go
Benehiko: also, O(n) means we are looping over all plugins, we will always need to loop over it, regardless of if it returns instantly or not.

joe0BAB: Right, I'll update my comment to `T(1)`, the time complexity is constant now, that's what this PR is about

pkg/adaptation/engine_test.go
```diff
@@ -0,0 +1,51 @@
...
+	}()
+	<-r3.wait
+	close(r3.done)
+	<-r1.wait
+	close(r1.done)
+	<-r2.wait
+	close(r2.done)
+	err := <-stopErr
+	assert.ErrorContains(t, err, "mock3 closed\nmock1 closed\nmock2 closed")
+}
```
joe0BAB: This test is racy. Unfortunately I don't see any other way to test this other than testing for what we actually want but with very extreme values: That the close time doesn't accumulate linearly with the number of plugins but approximately is capped to t_max.

pkg/adaptation/engine_test.go
```diff
@@ -0,0 +1,47 @@
...
+func (m *mockRuntime) Close() error {
+	time.Sleep(10 * time.Millisecond)
+	return fmt.Errorf("%s closed", m.name)
+}
+
+func (m *mockRuntime) Data() pluginData {
+	return pluginData{}
+}
+
+func Test_parallelStop(t *testing.T) {
```
joe0BAB, line 33: I'm not aware of any other way other than actually using timeouts to make this test not racy. However, if we use extreme values (if the logic were broken and the sleeps would accumulate, it would build up to minutes easily), we can still make this a reliable test.

pkg/adaptation/engine_test.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (m *mockRuntime) Data() pluginData {
+	return pluginData{}
+}
+
+// Unfortunately, there's no wait to test this reliably using channels.
+// However, if we push the values to their extremes by choosing a small individual sleep timeout per close
+// and a very large number of "runtime"s, we can still turn this into a reliable test.
+// -> if the logic were broken and the sleeps would add vs run in parallel, that test if broken would run really long
```
Benehiko:
```suggestion
// Unfortunately, there's no way to test this reliably using channels.
// We instead have a tiny sleep per mockRuntime.Close() with a larger global timeout in case the paralellStop function locks.
```

pkg/adaptation/engine_test.go
```diff
@@ -0,0 +1,51 @@
...
+type mockRuntime struct {
+	name string
+}
+
+func (m *mockRuntime) GetSecret(context.Context, secrets.Request) (secrets.Envelope, error) {
+	return secrets.Envelope{}, nil
+}
+
+func (m *mockRuntime) Close() error {
+	time.Sleep(10 * time.Millisecond)
```
Benehiko, line 23: perhaps choose a random number between 1-10ms for the timeout, making each Close variable (which would be more realistic)

pkg/adaptation/engine_test.go
```diff
@@ -0,0 +1,51 @@
...
+	stopErr := make(chan error)
+	go func() {
+		stopErr <- parallelStop(runtimes)
+	}()
+	select {
+	case err := <-stopErr:
+		assert.ErrorContains(t, err, "r24")
+		assert.ErrorContains(t, err, "r32")
+	case <-time.After(time.Second):
+		t.Fatal("timeout: parallel stop should not exceed 100ms")
```
Benehiko: 1 second > 100ms. 

```suggestion
	case <-time.After(time.Second):
		t.Fatal("timeout: parallel stop should not exceed 1s")
```

pkg/adaptation/engine_test.go
joe0BAB: I'd want this to reliably blow if something is broken

Benehiko (approved):

PR #86, opened 2025-07-07
feat(sdk): introduce onClose callback

joe0BAB:
Adds onClose callback that later can be used for cleanup/auto-restart logic.


pkg/adaptation/plugin_test.go
```diff
@@ -333,11 +344,30 @@ func runAsyncWithTimeout(ctx context.Context, run func(ctx context.Context) erro
 	return runErr, cancel
 }
 
+func onCloseCheck() (func(), func() error) {
+	onCloseCalled := make(chan struct{})
+	return func() {
+			close(onCloseCalled)
```
Benehiko, line 350: make this a `sync.Once` for the `close(onCloseCalled)`

pkg/adaptation/plugin_test.go
```diff
@@ -333,11 +345,33 @@ func runAsyncWithTimeout(ctx context.Context, run func(ctx context.Context) erro
 	return runErr, cancel
 }
 
+func onCloseCheck() (func(), func() error) {
```
Benehiko, line 347: could we add a test that calls the `cb` twice in case the sync.OnceFunc is removed or behavior changes?

joe0BAB: Should we? It would be writing a test for a test helper. I mean we could. But so far I didn't have this high quality standard for test helpers (eg there's also no mutex used to protect `m.closeCheck = checkOnClose` further down below).

Benehiko: i just want to make sure the `close` on the channel won't panic

Benehiko (approved):

PR #88, opened 2025-07-07
feat(sdk): monitored registration

joe0BAB:
Implements registration logic. There are multiple things that can go wrong during registration. Essentially, there's a race situation between adding the runtime/plugin to the registry and the plugin crashing/stopping and removing it from the registry so we don't have dead runtimes/plugins hanging around in the registry.

And that crashing/stopping might happen at any time, during registration or after at some point.


pkg/adaptation/internal/register/register_test.go
```diff
@@ -0,0 +1,143 @@
...
+			test: func(t *testing.T) {
+				l := func(func()) (runtime.Runtime, error) {
+					return &mockRuntime{}, nil
+				}
+				reg := &mockRegistry{}
+				registrator := newRegistrator(reg, l)
+				// We call onClose before doing register() to make this test reliable
+				// as there's no way to other way to assure onClose gets called at the right
+				// moment while register() is running
+				registrator.onClose()
```
joe0BAB: This is bothering me a bit. It feels a bit like cheating, but I couldn't find any other way.

pkg/adaptation/internal/register/register.go
```diff
@@ -0,0 +1,67 @@
...
+
+type registrator struct {
+	launched chan struct{}
+
+	onClose  func()
+	register func() error
+	remove   func()
+}
+
+func newRegistrator(reg Registry, launch Launcher) *registrator {
```
joe0BAB: I'm still on the fence if this could be written/structured better and more testable/readable.

pkg/adaptation/internal/register/register.go
```diff
@@ -0,0 +1,67 @@
...
+			return err
+		}
+		r.remove = sync.OnceFunc(rm)
+		return nil
+	})
+	return r
+}
+
+func Register(reg Registry, launch Launcher) error {
+	return newRegistrator(reg, launch).register()
```
joe0BAB: The whole putting things into an internal package dance serves this line essentially, making sure this is the only way the registration can be called, that `registrator` only can be used as an internal one time object, that exists for this line and so we can write better tests.

pkg/adaptation/internal/register/register.go
```diff
@@ -0,0 +1,73 @@
...
+	launched chan struct{}
+
+	onClose  func()
+	register func() error
+	remove   func()
+}
+
+func newRegistrator(reg Registry, launch Launcher) *registrator {
+	r := &registrator{
+		launched: make(chan struct{}),
```
Benehiko: Can't we make `launched` and atomic.Bool?

pkg/adaptation/internal/register/register.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	added := make(chan struct{})
+	onCloseCalled := make(chan struct{})
+	r.onClose = sync.OnceFunc(func() {
+		defer close(onCloseCalled)
+		select {
+		case <-r.launched:
```
Benehiko: Then we don't need this select with default.

pkg/adaptation/internal/register/register.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		rm, err := reg.Register(run)
+		if err != nil {
+			// TODO: Maybe we should send the shutdown reason to the plugin before shutting down?
+			if err := run.Close(); err != nil {
+				logrus.Error(err)
+			}
+			return err
+		}
+		r.remove = sync.OnceFunc(rm)
```
Benehiko: This is super confusing to me, so afaict this is almost like a context cancelFunc? 

Maybe we should change the interface to return a typed value instead of just a func(), similar to context's cancelFunc https://pkg.go.dev/context#CancelCauseFunc

pkg/adaptation/internal/register/register.go
joe0BAB: yes, initially I had this and then convinced myself we'd need that channel. But I think you're right, will change it back

pkg/adaptation/engine.go
```diff
@@ -24,3 +26,27 @@ func parallelStop(plugins []runtime) error {
 	}
 	return errs
 }
+
+type Launcher func() (runtime, <-chan struct{}, error)
```
joe0BAB, line 30: I'm cleaning this up further (moving the channel into the `runtime` interface) in #90 (didn't want to have to many different topics in this PR)

Benehiko (approved):
Looks much better! :partying_face:

PR #89, opened 2025-07-08
feat(sdk): runtime registry implementation

joe0BAB:
Implementation of a concurrency safe runtime registry.

For now this is very simple, it's just a sorted list. Later we might want to add additional checks (eg plugins not overwriting another plugins pattern).


pkg/adaptation/manager.go
```diff
@@ -1,8 +1,71 @@
...
+
+func (m *manager) Register(plugin runtime) (removeFunc, error) {
+	m.m.Lock()
+	defer m.m.Unlock()
+	if plugin.Data().name == "" {
+		return nil, errors.New("no plugin name")
+	}
+	for _, p := range m.plugins {
+		if p.Data().name == plugin.Data().name {
+			fmt.Printf("%s == %s\n", p.Data().name, plugin.Data().name)
```
Benehiko: Is this print for debugging purposes? Or can we instead log?

pkg/adaptation/manager.go
```diff
@@ -1,8 +1,71 @@
...
+	defer m.m.RUnlock()
+	return m.plugins
+}
+
+func (m *manager) remove(plugin runtime) {
+	m.m.Lock()
+	defer m.m.Unlock()
+	for i, p := range m.plugins {
+		if p == plugin {
+			m.plugins = append(m.plugins[:i], m.plugins[i+1:]...)
```
Benehiko, line 68: I'm wondering about this function a bit. Couldn't we store the plugins in a map instead? It would make lookup O(1) and Deletes O(1).

pkg/adaptation/manager.go
```diff
@@ -1,8 +1,71 @@
...
+	m.sort()
+	return func() {
+		m.remove(plugin)
+	}, nil
+}
+
+func (m *manager) sort() {
+	sort.Slice(m.plugins, func(i, j int) bool {
+		return m.plugins[i].Data().name < m.plugins[j].Data().name
+	})
```
Benehiko: Why would we sort before checking the size of the plugins? This sort is also not stable https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/sort/slice.go;l=24. 

I think we could rather keep track of the order of plugins in a separate array with only the name of the plugin as a value. Something like `orderedPlugins := []string{}`

pkg/adaptation/manager.go
joe0BAB: I'll delete this, it's a debug leftover

pkg/adaptation/manager.go
joe0BAB: While technically true, in reality insert/remove will happen very rarely so we basically don't care if insert/remove have O(n). Second, n here is expected to be <100, so it doesn't really matter anyway.

-> I'd prefer to keep this a simple slice for now and avoid what feels like premature optimization 

(However we do care that `GetAll` is O(1) )

pkg/adaptation/manager.go
joe0BAB: While `sort.Slice` might be unstable in general, in our case it's stable as plugin names are guaranteed to be distinct.

See comment about pre-mature optimization why I'd prefer sticking with a slice for now.

pkg/adaptation/manager.go
Benehiko: I'm still more convinced about the map here, it will reduce a lot of the code you have here and in addition give you O(1) - even if the num of plugins are small.

The code does a lot of loops over plugin just to find if it exists, a simple map[key] lookup would remove the need for the loops in line 65, and line 32. Do you need the plugins to be sorted (e.g. have it alphabetically sorted)?

pkg/adaptation/manager.go
Benehiko: Then you should use `slices.SortFunc` instead https://pkg.go.dev/slices#SortFunc as commented here https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/sort/slice.go;l=22

As commented below, do we really need a sorted slice for plugin names?

joe0BAB: I wanted to keep behavior as is for now, ie port https://github.com/docker/secrets-engine/blob/299adf0813d7e7234c9d6c78f7c5fbcc60443f20/pkg/engine/engine.go#L44 which picks the first plugin. Then it's better to have a defined order.

In any case, I'll expect the implementation of `manager` to change once more requirements come in. For now I need something simple that aligns with sjd's initial implementation and gives me O(1) on `GetAll`

joe0BAB: we can switch to `slices.SortFunc` though, I'll push a change

Benehiko (approved):

joe0BAB:
Note: We'll revisit this once have discussed the secret resolution requirements

PR #91, opened 2025-07-08
feat(sdk): plugin discovery scan

joe0BAB:
Basically scan a given directory for binary files and ignore everything else plus some tiny error handling / robustness.


pkg/adaptation/engine.go
```diff
@@ -50,3 +53,38 @@ func register(reg registry, launch Launcher) error {
...
 }
+
+func discoverPlugins(pluginPath string) ([]string, error) {
+	var result []string
+
+	entries, err := os.ReadDir(pluginPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			logrus.Warnf("Plugin directory does not exist: %s", pluginPath)
+			return nil, nil
```
joe0BAB, line 68: This could be debated. I think for now it's ok as is, however we might want to revert this later. I'm not really sure what's good behavior if someone deleted the plugin dir, fail fast or ignore?

Benehiko (approved):

PR #92, opened 2025-07-08
feat(sdk): parallelized start up logic

joe0BAB:
Startup logic to launch all plugins that are discovered in the plugin directory.


pkg/adaptation/plugin_test.go
```diff
@@ -70,11 +68,23 @@ func (m mockedPlugin) Config() p.Config {
...
 // TestMain acts as a dispatcher to run as dummy plugin or normal test.
 // Inspired by: https://github.com/golang/go/blob/15d9fe43d648764d41a88c75889c84df5e580930/src/os/exec/exec_test.go#L69-L73
 func TestMain(m *testing.M) {
-	if os.Getenv("RUN_AS_DUMMY_PLUGIN") != "" {
-		dummyPluginProcess()
+	binaryName := getTestBinaryName()
+	if strings.HasPrefix(binaryName, "plugin") {
+		// This allows tests to call the test binary as plugin by creating a symlink to it
+		// -> if the symlinked binary name matters a pattern it runs as plugin instead of a normal test
+		dummyPluginProcessFromBinaryName(binaryName)
```
Benehiko: is it necessary to symlink? Could we instead accept an `io/fs` for plugin discovery and in the test code use https://pkg.go.dev/testing/fstest#MapFS to create a fake directory with known plugins?

Benehiko: The comment also doesn't read well, line 84 makes little sense.

joe0BAB: I'll rewrite it, that's indeed gibberish 

regarding above: yes, we want to run/test this cross platform soon

Benehiko (approved):

PR #93, opened 2025-07-09
feat(sdk): client module

### LINKED ISSUE #94: Client SDK
The client SDK is the official way to communicate with the engine. It's a Go module that can be imported by whoever needs to communicate with the engine.

joe0BAB:
Client SDK module that can be imported and is the official SDK to communicate with the engine.

Note: This is initial plumbing work. Not all fields are supported yet by the IPC and engine. This will be added in follow up PRs step by step.


client/client.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return api.EnvelopeErr(request, err), err
+	}
+	id, err := secrets.ParseID(resp.Msg.GetSecretId())
+	if err != nil {
+		return api.EnvelopeErr(request, err), err
+	}
+	return secrets.Envelope{
+		ID:    id,
+		Value: []byte(resp.Msg.GetSecretValue()),
```
joe0BAB, line 89: There are obviously fields missing here but we don't have them in the protobuf yet.

Benehiko (approved):

PR #95, opened 2025-07-09
feat(sdk): secrets engine

joe0BAB:
A minimal secrets engine.


pkg/adaptation/engine_test.go
```diff
@@ -186,9 +184,36 @@ func Test_startPlugins(t *testing.T) {
...
 	assert.Len(t, plugins, 1)
-	assert.Equal(t, dummyPluginOk, plugins[0].Data().name)
+	assert.Equal(t, okPlugin, plugins[0].Data().name)
 	for _, plugin := range plugins {
 		assert.NoError(t, plugin.Close())
 	}
 	assert.Empty(t, reg.GetAll())
 }
+
+func Test_newEngine(t *testing.T) {
```
joe0BAB, line 194: I'll expand that test in follow up PRs.

pkg/adaptation/dummy.go
```diff
@@ -9,41 +9,72 @@ import (
...
 
+type dummyPlugins struct {
+	failPlugin bool
+	okPlugins  []string
+}
+
+// Use it in a test to create a set of dummy plugins that behave like normal plugins
+// but under the hood re-use the test binary.
+// This is the counterpart to dummyPluginProcessFromBinaryName().
+func createDummyPlugins(t *testing.T, cfg dummyPlugins) string {
```
Benehiko, line 64: i think here is where we would use t.Helper()

pkg/adaptation/engine.go
```diff
@@ -1,19 +1,69 @@
...
+			defer l.Close()
+			select {
+			case <-done:
+				return serverErr
+			default:
+			}
+			stopErr := parallelStop(reg.GetAll())
+			ctx, cancel := context.WithTimeout(context.Background(), engineShutdownTimeout)
+			defer cancel()
+			return errors.Join(server.Shutdown(ctx), stopErr)
```
Benehiko: I think here is a missing edge case. 
- I close server
- it gives me back an error
- It gets ignored with the select on line 50 (since server is not closed yet)
- falls through and eventually executes `defer l.Close()`.

Basically in some cases we are ignoring `serverErr`.

Benehiko: perhaps instead of managing a `done` channel and a `serverErr`, make `serverErr` a channel instead

pkg/adaptation/dummy.go
joe0BAB: yes!

pkg/adaptation/engine.go
joe0BAB: Ignoring the server error value on explicit shutdown is intentional as the only errors it returns in this case is io.EOF or ErrServerShutdown. But there's a different bug: We wouldn't shutdown the plugins. I'll fix that.

And to still address your point: Maybe we should include server error but filter out io.EOF and ErrServerShutdown? (we right now kinda rely on that I know the current implementation)

pkg/adaptation/engine.go
```diff
@@ -1,19 +1,77 @@
...
+	var serverErr error
+	go func() {
+		err := server.Serve(l)
+		stopErr := stopPlugins()
+		// Only forward errors from here when the server crashed,
+		// but not when it's explicitly being shutdown
+		// -> don't double report stopPlugins() error and io.EOF/ErrServerClosed just mean we have shutdown
+		if !errors.Is(err, http.ErrServerClosed) && !errors.Is(err, io.EOF) {
+			serverErr = errors.Join(err, stopErr)
+		}
```
Benehiko: so `stopErr` will only be added to the `serverErr` if server crashes?

joe0BAB: Yes, as otherwise we'd add it twice in line 66 as stopPlugins would return the error from the previous stopPlugins due to the once wrapper, wouldn't it?

joe0BAB: Maybe this could be written more explicit?

Benehiko: it's not obvious to me that line 63 would be called again if server had stopped before close was called. In that case line 60 would return, leaving out any errors from `stopPlugins()`

joe0BAB: Rewrote it. Could you check again? 😅

pkg/adaptation/engine.go
```diff
@@ -1,19 +1,79 @@
...
+	server := newServer(reg)
+	done := make(chan struct{})
+	var serverErr error
+	go func() {
+		err := server.Serve(l)
+		if !errors.Is(err, http.ErrServerClosed) && !errors.Is(err, io.EOF) {
+			err = nil
+		}
+		serverErr = errors.Join(err, stopPlugins())
+		close(done)
```
Benehiko:
```suggestion
		if err := server.Serve(l); !errors.Is(err, http.ErrServerClosed) && !errors.Is(err, io.EOF) {
			serverErr = errors.Join(serverErr, err)
		}
		serverErr = errors.Join(serverErr, stopPlugins())
		close(done)
```

Benehiko (approved):

PR #98, opened 2025-07-10
ci(sdk): add CI pipeline to check engine code on mac

### LINKED ISSUE #34: IPC: Windows support

Image

```go
// parent.go
package main

import (
    "fmt"
    "os"
    "os/exec"
    "strconv"
    "syscall"
    "unsafe"
)

func main() {
    }
    }

    )

    }

    // Start the child:
    }

    buf := make([]byte, 128)
    n, err := rFile.Read(buf)
    if err != nil {
    }

    cmd.Wait()
}
```

and

```go
// child.go
package main

import (
    "fmt"
    "os"
    "strconv"
)

func main() {
    if hstr == "" {
        panic("PIPE_HANDLE not set")
    }
    if err != nil {
    }

    defer pipe.Close()

    msg := "hello from child\n"
    }
}
```

joe0BAB:
Adds a cross OS ci pipeline for the engine code. For now it's mac only. Windows will follow soon.


.github/workflows/engine.yml
```diff
@@ -0,0 +1,24 @@
+name: Engine
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
Benehiko, line 9: Maybe we could add a path check here for only engine code?

https://github.com/docker/secrets-engine/blob/main/.github/workflows/keychain.yml#L9-L10

Makefile
```diff
@@ -60,6 +60,9 @@ keychain-linux-unit-tests:
 keychain-unit-tests:
 	CGO_ENABLED=1 go test -v $$(go list ./store/keychain/...)
 
+engine-unit-tests:
+	CGO_ENABLED=1 go test -v $$(go list ./engine/...)
```
Benehiko: do we need cgo on engine?

Benehiko (approved):

PR #101, opened 2025-07-14
feat(sdk): make graceful shutdown work on windows

### LINKED ISSUE #34: IPC: Windows support

joe0BAB:
Implements graceful shutdown for windows.

Note: The old way to do this was https://github.com/golang/go/blob/master/src/os/signal/signal_windows_test.go

For follow up PR:

```
		CreationFlags: syscall.CREATE_NEW_PROCESS_GROUP,
	}
```
has to be set when launching plugins on windows.


engine/exec.go
```diff
@@ -53,7 +59,31 @@ func isSigint(err error) bool {
...
+	}()
+	if err := askProcessToStop(cmd); err != nil {
+		logrus.Errorf("sending SIGINT/CTRL_BREAK_EVENT to plugin: %v", err)
+	} else {
+		select {
+		case <-done:
+			return
+		case <-time.After(pluginShutdownTimeout):
+		}
+	}
```
Benehiko: when reading this, i got confused.

So the select will only fire if `askProcessToStop` succeeds, but `cmd.Process.Kill()` will always be called. Maybe we can refactor this a bit.

engine/exec.go
```diff
@@ -53,7 +59,31 @@ func isSigint(err error) bool {
...
+	if err := cmd.Process.Kill(); err != nil {
+		logrus.Errorf("sending SIGKILL to plugin: %v", err)
+	}
+	<-done // wait before calling Release()
```
Benehiko:
```suggestion
	err := askProcessToStop(cmd)
	
	if err == nil {
	    select {
	        case <-done:
		    return
	        case <-time.After(pluginShutdownTimeout):
	    }
	}
	
	logrus.Errorf("sending SIGINT/CTRL_BREAK_EVENT to plugin: %v", err)
	
	if err := cmd.Process.Kill(); err != nil {
		logrus.Errorf("sending SIGKILL to plugin: %v", err)
	}
	
	// I still don't know what else we can use here to prevent it from waiting forever (in case something goes wrong)
	select {
	    case <-done:
	    case <-cmd.Context().Done():
	}
	
	if err := cmd.Process.Release(); err != nil {
		logrus.Errorf("release process err: %v", err)
	}
```

engine/exec_unix.go
```diff
@@ -3,37 +3,10 @@
...
-const (
-	pluginShutdownTimeout = 2 * time.Second
 )
 
-func shutdownCMD(cmd *exec.Cmd, done chan struct{}) {
-	if cmd.Process == nil {
-		return
-	}
-	defer func() {
-		if err := cmd.Process.Release(); err != nil {
```
joe0BAB, line 22: I read the docs, we don't need to call that as we already call wait: https://pkg.go.dev/os#Process.Release 😅

Benehiko (approved):

PR #99, opened 2025-07-10
feat(sdk): windows support

joe0BAB:
Refactors the IPC to get rid of the dependency on NRI and add support for windows.

This closes #34 .

PR #107, opened 2025-07-14
tests: make plugin discovery test work on windows

### LINKED ISSUE #34: IPC: Windows support

joe0BAB:
More cleanup to get windows support working and tests green.

(preparation for #99 )


engine/dummy.go
```diff
@@ -68,15 +68,30 @@ func createDummyPlugins(t *testing.T, cfg dummyPlugins) string {
 	assert.NoError(t, err)
 	dir := t.TempDir()
 	if cfg.failPlugin {
-		assert.NoError(t, os.Symlink(exe, filepath.Join(dir, dummyPluginFail)))
+		assert.NoError(t, copyFile(exe, filepath.Join(dir, dummyPluginFail+suffix)))
 	}
 	for _, p := range cfg.okPlugins {
 		require.True(t, strings.HasPrefix(p, "plugin-"))
-		assert.NoError(t, os.Symlink(exe, filepath.Join(dir, p)))
+		assert.NoError(t, copyFile(exe, filepath.Join(dir, p+suffix)))
```
joe0BAB, line 76: Not sure why, but symlinking on win gives permission issues -> we just copy instead

engine/dummy.go
```diff
@@ -68,15 +68,30 @@ func createDummyPlugins(t *testing.T, cfg dummyPlugins) string {
...
 
+func copyFile(src, dst string) error {
+	data, err := os.ReadFile(src)
+	if err != nil {
+		return &os.PathError{Op: "read", Path: src, Err: err}
+	}
+	info, err := os.Stat(src)
+	if err != nil {
+		return &os.PathError{Op: "stat", Path: src, Err: err}
+	}
```
Benehiko: why are we reading before `os.Stat`? Wouldn't it make sense to stat and then check if it exists?

engine/engine_unix.go
```diff
@@ -7,6 +7,12 @@ import (
 	"os"
 )
 
+const suffix = ""
```
Benehiko, line 10: suffix is unused i think

engine/dummy.go
joe0BAB: it doesn't really matter if os.ReadFile or os.Stat would fail first with a "not exist" error. Yes, typically you'd use os.Stat to check for existence, but os.ReadFile does that check as well. Here we just need to read both, content and meta data but I don't see why the order should make any difference.

engine/engine_unix.go
joe0BAB: Don't we have a linter to check that? 😂 

(spoiler: it is being used in a file without OS specific suffix)

Benehiko: it didn't seem obvious to me that you were using it in the tests and dummy plugin code. I suppose the linter would've caught it

engine/dummy.go
Benehiko: Yes, but then we can rather combine them by using `os.Open` and just return its error instead of creating a new `os.PathError`.

```go
func copyFile(src, dst string) error {
	f, err := os.Open(src)
	if err != nil {
		return err
	}
	defer f.Close()

	info, err := f.Stat()
	if err != nil {
		return err
	}

	data, err := io.ReadAll(f)
	if err != nil {
		return err
	}

	if err := os.WriteFile(dst, data, info.Mode().Perm()); err != nil {
		return &os.PathError{Op: "write", Path: dst, Err: err}
	}
	return nil
}
```

joe0BAB: Nice! I'll change to that!

engine/dummy.go
```diff
@@ -68,15 +69,38 @@ func createDummyPlugins(t *testing.T, cfg dummyPlugins) string {
...
-		assert.NoError(t, os.Symlink(exe, filepath.Join(dir, p)))
+		assert.NoError(t, copyFile(exe, filepath.Join(dir, p+suffix)))
 	}
 	return dir
 }
 
+func copyFile(src, dst string) error {
+	f, err := os.Open(src)
+	if err != nil {
+		return &os.PathError{Op: "open", Path: src, Err: err}
```
Benehiko: don't we already get os.PathError from `os.Open()`? https://pkg.go.dev/os#Open

I don't think it's necessary to wrap them again

joe0BAB: indeed, they all do, will remove

Benehiko (approved):

PR #108, opened 2025-07-14
store: add Filter function

### LINKED ISSUE #30: Store should support partial searches
The `Store` interface should support a `Search` implementation for partial searches. 
https://github.com/docker/secrets-engine/blob/main/store/store.go#L46-L55

For example, I have a secret ID with the format `realm/group/service/username`. I would like to be able to partially search for all items that are under the `group` key or all items under the `service` key.

Most of the keychain providers seem to support a partial search via `attributes`.

This issue proposes a new `Search` function that accepts a map of attributes:

```go
func Search(ctx context.Context, items map[string]string) ([]Secrets, error)
```

Benehiko:
closes #30 

The `Store.GetAll` function is poorly optimized, since it fetches all secrets for the current application from the underlying keychain. This unfortunately would lead to the user being prompted on macOS when each secret is decrypted. The secrets engine has only one option to fetch all secrets from the keychain on each provider (potentially with slight optimization to only do so based on the secret ID matching a certain provider). However, this would still be extremely inefficient. (fixed in https://github.com/docker/secrets-engine/pull/130)

~~This patch changes the behavior of `Store.GetAll` by only retrieving the secrets metadata from the underlying store, which is a non-sensitive set of attributes that requires no password prompt from the user.~~ Done in https://github.com/docker/secrets-engine/pull/130. It further adds an additional function called `Store.Filter` to fetch secrets from the store based on certain criteria. This allows a store provider to filter secrets based on partial IDs and attributes.

Below is an example of a new function that can accept a Pattern. It produces a populated map of secrets that could be one or many. On zero results it would return an error.

```go
func Filter(ctx context.Context, pattern Pattern) (map[string]Secret, error)
```

Usages would be as follows:

```go
// get all secrets from the store that has an ID starting with `com.test.group`.
Filter(ctx, store.MustParsePattern("com.test.group/**"))
```

We can even filter based on a Pattern where anything can be before and after the match:

```go
// get all secrets where all attributes below match under the partial ID `com.test.group`. 
Filter(ctx, store.MustParsePattern("**/app-name/**"))
```

---

requires ~https://github.com/keybase/go-keychain/pull/113~

~~Changed from https://github.com/keybase/go-keychain to https://github.com/Benehiko/go-keychain/v2 so that we don't need to wait for any upstream merges. I've added additional commits to my fork so we can use the more up to date go dbus library https://github.com/godbus/dbus.~~ Done in https://github.com/docker/secrets-engine/pull/128.


store/mocks/mock_store.go
```diff
@@ -2,64 +2,113 @@ package mocks
...
+
+	filtered := make(map[store.ID]store.Secret)
+	for _, f := range m.store {
+		for _, p := range parts {
+			if vv, ok := f.attributes[p]; ok && p == vv {
+				filtered[f.id] = f.store
+				continue
+			}
+		}
+		for k, v := range attributes {
```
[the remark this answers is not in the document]
Benehiko: no...

store/mocks/mock_store.go
```diff
@@ -13,53 +14,56 @@ type MockStore struct {
...
+
+	filtered := make(map[store.ID]store.Secret)
+	for id, f := range m.store {
+		for _, p := range parts {
+			if vv, ok := f.Metadata()[p]; ok && p == vv {
+				filtered[id] = f
+				continue
+			}
+		}
+		for k, v := range attributes {
```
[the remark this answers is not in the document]
Benehiko: no it wont.

store/store.go
Benehiko, line 1: This is the important part of the change. We are adding to the `Store` and `Secret` interface and changing the behavior of the `Store.GetAll` function.

store/keychain/cmd/main.go
```diff
@@ -64,15 +63,25 @@ func newCommand() (*cobra.Command, error) {
 			if err != nil {
 				return err
 			}
+
+			secretMetadata := make(map[string]string)
+			for _, vals := range metadata {
+				vv := strings.Split(vals, ":")
```
joe0BAB: Maybe we should reflect this in the command help text?

joe0BAB: This implicit behavior in the next line I mean

store/keychain/keychain_darwin.go
```diff
@@ -143,9 +165,97 @@ func (k *keychainStore[T]) Save(_ context.Context, id store.ID, secret store.Sec
...
 }
 
+func hasAttribute(searchAttributes, queryAttributes map[string]string) bool {
+	if searchAttributes == nil {
+		return true
+	}
+	allMatches := make([]bool, len(searchAttributes))
+	var i int
+	for k, v := range queryAttributes {
+		if needle, ok := searchAttributes[k]; ok && strings.EqualFold(needle, v) {
```
joe0BAB: Why can't we reverse this and early return with false? I don't think we need that `allMatches`

store/keychain/keychain_test.go
```diff
@@ -112,17 +131,85 @@ func TestKeychain(t *testing.T) {
 		}
 		secrets, err := ks.GetAll(t.Context())
 		require.NoError(t, err)
+		require.Len(t, secrets, 3)
 
-		actual := make(map[store.ID]*mocks.MockCredential)
-		for id, s := range secrets {
-			actual[id] = s.(*mocks.MockCredential)
+		expected := make([]store.ID, 0, 3)
```
joe0BAB: Can we keep this boring and just write `var expected []store.ID`?

store/keychain/keychain_test.go
```diff
@@ -112,17 +131,85 @@ func TestKeychain(t *testing.T) {
...
+		for k := range moreCreds {
+			expected = append(expected, k)
 		}
-		require.Len(t, actual, 3)
+		slices.Sort(expected)
 
-		expected := moreCreds
+		actual := make([]store.ID, 0, 3)
```
joe0BAB: same, let's keep this boring

store/keychain/keychain_test.go
```diff
@@ -112,17 +131,85 @@ func TestKeychain(t *testing.T) {
...
+		for key := range secrets {
```
joe0BAB: Can we replace that with maps.Keys from golang.org/x/exp/maps?

store/keychain/keychain_test.go
```diff
@@ -112,17 +131,85 @@ func TestKeychain(t *testing.T) {
 		}
 		secrets, err := ks.GetAll(t.Context())
 		require.NoError(t, err)
+		require.Len(t, secrets, 3)
 
-		actual := make(map[store.ID]*mocks.MockCredential)
-		for id, s := range secrets {
-			actual[id] = s.(*mocks.MockCredential)
+		expected := make([]store.ID, 0, 3)
+		for k := range moreCreds {
```
joe0BAB: Can we use maps.Keys here?

store/keychain/keychain_test.go
```diff
@@ -112,17 +131,85 @@ func TestKeychain(t *testing.T) {
...
 		}
-		require.Len(t, actual, 3)
+		slices.Sort(expected)
 
-		expected := moreCreds
+		actual := make([]store.ID, 0, 3)
+		for key := range secrets {
+			actual = append(actual, key)
+		}
+		slices.Sort(actual)
```
joe0BAB: What about using `assert.ElementsMatch`?

store/keychain/keychain_test.go
```diff
@@ -112,17 +131,85 @@ func TestKeychain(t *testing.T) {
...
 		require.Equal(t, expected, actual)
```
joe0BAB: Is there a reason we only compare the keys but not the entire map?

store/keychain/keychain_test.go
```diff
@@ -2,6 +2,7 @@ package keychain
 
 import (
 	"errors"
+	"slices"
 	"testing"
 
 	"github.com/stretchr/testify/require"
```
joe0BAB: Are we intentionally using require here? I feel most of the time you want to use assert unless continuing the test would cause a panic.

store/keychain/keychain_darwin.go
Benehiko: We use the `allMatches` so that we check if all attributes we are searching for were found in the attributes retrieved from the keychain. 

That means the behavior of filter is not an OR match it's an AND match. 

```
attributes = { "color": "green", "sport", "elden ring" }
filter = {"color" : "yellow", "music": "classical", "sport": "elden ring"}
```

In the above scenario I would not get a match, because "color" is not "yellow" and I have no field "music", so we skip, even though we fulfill the match `"sport": "elden ring"`.

So something like `SELECT * FROM keychain WHERE values @> ARRAY[1, 3, 4]`

store/keychain/keychain_test.go
Benehiko: require is quite good, but i agree, we should use assert when checking the values coming back from the keychain. I'll update it a bit.

store/keychain/keychain_darwin.go
```diff
@@ -111,16 +135,14 @@ func (k *keychainStore[T]) GetAll(context.Context) (map[store.ID]store.Secret, e
 	for _, result := range results {
 		id, err := store.ParseID(result.Account)
 		if err != nil {
-			return nil, err
+			continue
```
joe0BAB: Why is this no longer an error? 🤔

Benehiko: i think returning an error here is incorrect. We might overlap with "non-conforming" secrets inside the keychain. And if we were to error on them, the others won't be retrieved.

store/keychain/keychain_darwin.go
joe0BAB: > That means the behavior of filter is not an OR match it's an AND match.

That's exactly what I mean. An early return `false` should give you that. Also we should add a unit test for this function.

store/keychain/keychain_darwin.go
```diff
@@ -143,9 +165,97 @@ func (k *keychainStore[T]) Save(_ context.Context, id store.ID, secret store.Sec
...
+		}
+		metadata[p] = p
+	}
+
+	item.SetGenericMetadata(metadata)
+
 	return mapKeychainError(kc.AddItem(item))
 }
 
+func hasAttribute(searchAttributes, queryAttributes map[string]string) bool {
```
joe0BAB: To me this name is confusing. It's singular and yet we pass in two plural arguments.

store/keychain/keychain_darwin.go
Benehiko: how can you early return when `queryAttributes` might be longer in length than `searchAttributes`. I could try see to optimise the code here. In some cases that's not possible, like in Windows (since it returns a list and not a map).

store/store.go
```diff
@@ -49,7 +55,27 @@ type Store interface {
...
 	Get(ctx context.Context, id ID) (Secret, error)
 	// GetAll retrieves all the credentials from the store.
+	// Credentials retrieved will only call [Secret.SetMetadata] so that the
+	// underlying store does not get queried for each secret's sensitive data.
+	// This could be very taxing on the underlying store and cause a poor User
+	// Experience.
 	GetAll(ctx context.Context) (map[ID]Secret, error)
 	// Save persists credentials from the store.
 	Save(ctx context.Context, id ID, secret Secret) error
+	// Filter returns a map of secrets based on a partial ID and metadata.
```
joe0BAB: Hm, I'm having mixed feelings about this "partial ID" concept. Why not re-use patterns here? I would have expected and interface like this:

`Filter(ctx, pattern, attributes)`

And pattern of course can be an exact ID. That would allow us to re-use `pattern.Match`/`id.Match` logic and we'd remain consistent / not have to come up with another concept of partial IDs here.

store/keychain/keychain_linux.go
```diff
@@ -318,6 +319,17 @@ func (k *keychainStore[T]) Save(_ context.Context, id store.ID, secret store.Sec
 	}
 
 	attributes := newItemAttributes(id.String(), k)
+
+	parts := strings.SplitSeq(id.String(), "/")
+	for p := range parts {
+		if p == "" {
+			continue
+		}
+		attributes[p] = p
```
joe0BAB: This scares me. So I assume you breakdown a path into its components as on a lower level secrets on linux cannot be filtered via path but via attributes only I assume?

But then, you don't protect these attributes later for `maps.Copy(attributes, secret.Metadata())` which could completely break this?

Eg `ID = "/foo/bar"` with attributes `bar=baz`. But then also it might work as for the first you could only rely on the keys and not the values?

Either way, could we get some tests and comments around this? 😅

store/store.go
Benehiko: The underlying store might not support our pattern, so we'd have to somehow convert it. 

Partial ID just means you can search by any sub-set of the ID, e.g. "com.test.test/applicationID/resourceURL/username" -> "com.test.test" is valid, "com.test.test/applicationID" is valid, "com.test.test/applicationID/resourceURL" is valid.

store/keychain/keychain_linux.go
joe0BAB: (a unit test for that algorithm only could be enough I think)

store/keychain/keychain_linux.go
```diff
@@ -318,6 +319,17 @@ func (k *keychainStore[T]) Save(_ context.Context, id store.ID, secret store.Sec
 	}
 
 	attributes := newItemAttributes(id.String(), k)
+
+	parts := strings.SplitSeq(id.String(), "/")
```
joe0BAB: I'm wondering if we should add a `Components` method to `ID`. It bothers me a bit that we rely on the implementation detail of `/` here. (we could also make `const Separator = "/"` an exported constant and then import it here and just have a local function `components(ID) []string` in this file)

store/keychain/keychain_linux.go
```diff
@@ -328,3 +340,87 @@ func (k *keychainStore[T]) Save(_ context.Context, id store.ID, secret store.Sec
...
+	itemPaths, err := service.SearchCollection(objectPath, itemAttr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to search collection: %w", err)
+	}
+
+	if len(itemPaths) == 0 {
+		return nil, store.ErrCredentialNotFound
+	}
+
+	credentials := make(map[store.ID]store.Secret, len(itemPaths))
```
joe0BAB: I'd argue we're are fare of from that level of performance optimization here where we need to pre-allocate maps/slices.

store/keychain/keychain_linux.go
joe0BAB: Actually I think it's a bug and breaks here: `itemPaths, err := service.SearchCollection(objectPath, itemAttr)`

store/keychain/keychain_linux.go
```diff
@@ -328,3 +340,87 @@ func (k *keychainStore[T]) Save(_ context.Context, id store.ID, secret store.Sec
...
+
+	maps.Copy(itemAttr, attributes)
+
+	parts := strings.SplitSeq(id.String(), "/")
+	for p := range parts {
+		if p == "" {
+			continue
+		}
+		itemAttr[p] = p
+	}
```
joe0BAB: similar as above, I suggest to wrap this into eg a `convertPathToSearchAttributes` or something function

store/keychain/keychain_darwin.go
joe0BAB: might deserve a comment in the code

store/store.go
joe0BAB: But couldn't we write this as a pattern? `com.test.test/**`?

joe0BAB: It sounds like "partial ID" is a subclass of pattern. However, it's not very explicit defined but very implicit through convention. Maybe we could change that.

joe0BAB: It sounds like "partial ID" is a subclass of pattern. However, it's not very explicit defined but very implicit through convention. Maybe we could change that.

store/keychain/keychain.go
```diff
@@ -67,30 +67,30 @@ const (
 	secretIDKey     = "id"
 )
 
-// safelySetMetadata prefixes each key with `x_` so that no collissions can ever
-// occur with internal fields.
-func (k *keychainStore[T]) safelySetMetadata(id string, attributes map[string]string) {
-	// prefix whatever is already in attributes
+// safelySetMetadata is a helper function to keychain providers
+// it adds internal metadata as well as prefixes externally defined attribute
```
joe0BAB, line 71:
```suggestion
// safelySetMetadata is a helper function to keychain providers.
// It adds internal metadata as well as prefixes externally defined attribute
```

joe0BAB (approved):

PR #110, opened 2025-07-15
refactor: remove acceptPattern field and use better name

### LINKED ISSUE #66: Adaptation: Internal Plugins
Image

This complements the plugin system to also support "internal plugins", i.e., plugins that get shipped with the secrets engine by default (batteries included). This will be required e.g. for the auth and local secrets storage plugins bundled with Docker Desktop or docker-ce.

joe0BAB:
Came across this when working on the internal plugins logic. This is a leftover and was based on early assumptions that we had to refine recently. Also added this to the features list (unchecked and doesn't necessarily mean that we have to implement it but to keep track of it).


engine/setup.go
```diff
@@ -20,15 +19,14 @@ type setupResult struct {
 	close  func() error
 }
 
-var _ pluginCfgInValidator = &setupValidator{}
+var _ pluginCfgInValidator = &runtimeCfg{}
 
-type setupValidator struct {
-	out           pluginCfgOut
-	name          string
-	acceptPattern func(secrets.Pattern) error
```
joe0BAB, line 28: Now that we have the engine code, having this here makes little sense anymore. This validation needs to happen in the plugin registry code once we get to actually implement it.

Benehiko (approved):

PR #111, opened 2025-07-16
fix(ci): Dockerfile magic

joe0BAB:
This fixes CI:
- run all unit tests and return an error if any errors
- lint the entire codebase
- avoid downloading vendored dependencies


Dockerfile
```diff
@@ -41,15 +41,20 @@ COPY --link --from=lint-base /usr/bin/golangci-lint /usr/bin/golangci-lint
...
     (cd client && go mod tidy --diff)
     (cd engine && go mod tidy --diff)
     (cd plugin && go mod tidy --diff)
     (cd store && go mod tidy --diff)
-    golangci-lint run -v
 EOT
+RUN --mount=type=bind,target=.,ro \
+    --mount=type=cache,target=/root/.cache <<EOT
+    set -euo pipefail
+    golangci-lint run -v $(go list -f '{{.Dir}}/...' -m | xargs)
```
Benehiko, line 57: is this really necessary? I thought `golangci-lint run` already traverses directories recursively?

Makefile
```diff
@@ -52,11 +52,19 @@ clean: ## remove built binaries and packages
 	@sh -c "rm -rf bin dist"
 
 unit-tests:
-	go test -v $$(go list ./client/...) & \
-	go test -v $$(go list ./engine/...) & \
-	go test -v $$(go list ./plugin/...) & \
-	go test -v $$(go list ./...)      & \
-	wait
```
Benehiko, line 59: so parallel doesn't work?

Dockerfile
```diff
@@ -41,15 +41,20 @@ COPY --link --from=lint-base /usr/bin/golangci-lint /usr/bin/golangci-lint
...
     set -euo pipefail
     go mod tidy --diff
     (cd client && go mod tidy --diff)
     (cd engine && go mod tidy --diff)
     (cd plugin && go mod tidy --diff)
     (cd store && go mod tidy --diff)
-    golangci-lint run -v
 EOT
+RUN --mount=type=bind,target=.,ro \
+    --mount=type=cache,target=/root/.cache <<EOT
```
Benehiko, line 53: also we should then also change this `/root/.cache` if the above changes allowed caching go mod files.

Makefile
joe0BAB: it does, but only grabs the error of the last command. We need any error which requires a bit of shell shenanigans (Feel free to improve that part or suggest a better solution. I don't like it but it's the best I found)

Dockerfile
Benehiko: let's add a reference to this maybe https://github.com/golangci/golangci-lint/issues/2654#issuecomment-1606439587.

Makefile
Benehiko: alright, let's leave it as is and then improve later if we get time

Benehiko (approved):

PR #114, opened 2025-07-16
feat(sdk): runtime for internal plugins

### LINKED ISSUE #66: Adaptation: Internal Plugins

joe0BAB:
Implements the runtime interface for internal plugins.

Regarding internal plugins: They are similar to external plugins plus we always expect the worst that everything they return could be wrong or result in a panic.


engine/builtin.go
```diff
@@ -0,0 +1,118 @@
...
+	default:
+	}
+	defer func() {
+		if r := recover(); r != nil {
+			panicErr := fmt.Errorf("recovering from panic in plugin %s: %s", i.name, debug.Stack())
+			resp = secrets.EnvelopeErr(request, panicErr)
+			err = panicErr
+		}
+	}()
+	return i.p.GetSecret(ctx, request)
```
joe0BAB, line 100: We just forward here, the engine later will apply additional deeper checks on the values.

Benehiko (approved):

PR #117, opened 2025-07-16
tests: parallelize more tests

joe0BAB:


README.md
```diff
@@ -1,103 +1,95 @@
...
+## Installation
 
-The best way to get a handle on this is to compile the plugin:
-
-```console
-go build ./cmd/docker-secrets/
+```bash
+go get github.com/docker/secrets-engine/client
+go get github.com/docker/secrets-engine/engine
+go get github.com/docker/secrets-engine/plugin
```
Benehiko: hmm, i'm not sure. Installation should use `go install` instead which I think is possible for the engine. But the others are not singular binaries and should instead have a dedicated section along with the `go get` command.

README.md
```diff
@@ -1,103 +1,95 @@
...
 ```
 
-You can move it into the plugin location but its not necessary (and may not
-work).
+## Usage
 
-To start, create a few secrets in your keychain:
+### Client
```
Benehiko, line 10: add the `go get ...` for client here.

README.md
```diff
@@ -1,103 +1,95 @@
...
+if err != nil {
+    log.Fatalf("failed fetching secret: %v", err)
+}
+fmt.Println(resp.Value)
 ```
 
-In the above example, we can see both the file-based secret, which is the
-default and the explicit env secret. Selecting these secrets also makes them
-available via the API but only for the secrets we selected:
+### Writing your own Plugin
```
Benehiko, line 34: add the `go get ...` for plugin here.

README.md
```diff
@@ -1,103 +1,95 @@
...
 
-```console
-root@c1d2b75b4bb3:/# apt update -q && apt install -y -q curl jq
-...
-root@c1d2b75b4bb3:/# curl --unix-socket $SECRETS_SOCK http://localhost/secrets/resolve?id=foo/bar
-{"Secrets":[{"ID":"foo/bar","Value":"YXNkZgo=","Provider":"local","CreatedAt":"2025-05-05T17:55:36-07:00","ResolvedAt":"2025-05-13T02:29:43.221382Z"}]}
-```
+Any plugin needs to implement the `Plugin` interface:
```
Benehiko:
```suggestion
A plugin needs to implement the `Plugin` interface:
```

Benehiko (approved):

PR #120, opened 2025-07-17
feat(ipc): hijacking mechanics

### LINKED ISSUE #119: Engine: Integrate runtime for external plugins
We have a runtime already for external plugins, however, it still needs to be integrated with the engine.

joe0BAB:
Implements the basic mechanics of connection hijacking required to integrate the runtime for externally launched plugins into the engine.


internal/ipc/hijack.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (hcw *hijackedConnCloseWriter) CloseWrite() error {
+	if cw, ok := hcw.Conn.(CloseWriter); ok {
+		return cw.CloseWrite()
+	}
+	return nil
+}
+
+func Hijackify(conn net.Conn) (net.Conn, error) {
```
Benehiko: We should add a godoc here to indicate where you'd use it and how it is used.

internal/ipc/hijack.go
```diff
@@ -0,0 +1,143 @@
...
+		conn.Close()
+		return nil, fmt.Errorf("upgrading: %s", resp.Status)
+	}
+
+	if br.Buffered() > 0 {
+		if _, ok := conn.(CloseWriter); ok {
+			conn = &hijackedConnCloseWriter{&hijackedConn{conn, br}}
+		} else {
+			conn = &hijackedConn{conn, br}
+		}
```
Benehiko: no sure if this if statement is necessary. One could just always set `hijackedConnCloseWriter` and inside its function check if the underlying conn is a `CloseWriter`.

internal/ipc/hijack.go
```diff
@@ -0,0 +1,143 @@
...
+		conn.Close()
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusSwitchingProtocols {
+		conn.Close()
+		return nil, fmt.Errorf("upgrading: %s", resp.Status)
+	}
+
+	if br.Buffered() > 0 {
```
Benehiko: A bit confused by this check and why it should influence if we overwrite conn with `hijackedConnCloseWriter`. I also don't understand why `br.Reset(nil)` is only called if buffered is empty.

internal/ipc/hijack.go
```diff
@@ -0,0 +1,143 @@
...
+	conn, _, err := hj.Hijack()
+	if err != nil {
+		logrus.Errorf("hijack error: %v", err)
+		return
+	}
+
+	// Flush the options to make sure the client sets the raw mode
+	_, _ = conn.Write([]byte{})
+	if _, ok := r.Header["Upgrade"]; ok {
+		fmt.Fprint(conn, "HTTP/1.1 101 UPGRADED\r\nConnection: Upgrade\r\nUpgrade: tcp\r\n")
```
Benehiko: I couldn't find `UPGRADED` as a valid HTTP response status

101 is `switching protocols` 

You can see the Go std `reverseproxy` test code doing the same but using `101 Switching Protocols`. https://cs.opensource.google/go/go/+/master:src/net/http/httputil/reverseproxy_test.go;l=1651?q=UPGRADED&ss=go%2Fgo

Benehiko: for readability we could even re-use `http.Header{}`.

```go
headers := http.Header{
    "Connection": []string{"Upgrade"},
    "Upgrade":    []string{"tcp"},
}
var b bytes.Buffer
headers.Write(&b)
fmt.Println(b.String())
```

internal/ipc/hijack.go
```diff
@@ -0,0 +1,143 @@
...
+	} else {
+		fmt.Fprint(conn, "HTTP/1.1 200 OK\r\n")
+	}
```
Benehiko: I think one could even re-use the http.ResponseWriter since we are already using it on line 120.

```suggestion
       if _, ok := r.Header["Upgrade"]; ok {
		w.Header().Set("Connection", "Upgrade")
		w.Header().Set("Upgrade", "tcp")
		w.WriteHeader(http.StatusSwitchingProtocols)
	} else {
		w.WriteHeader(http.StatusOK)
	}
```

Benehiko: this just needs confirmation, unsure if "raw mode" has different behaviors.

internal/ipc/hijack.go
```diff
@@ -0,0 +1,143 @@
...
+		return
+	}
+	conn, _, err := hj.Hijack()
+	if err != nil {
+		logrus.Errorf("hijack error: %v", err)
+		return
+	}
+
+	// Flush the options to make sure the client sets the raw mode
+	_, _ = conn.Write([]byte{})
```
Benehiko: We might have access to `Flusher` already. 

on line 106, we might be able to get back the writer which has a `Flush` function. See this code inside the Go std for reverse proxy https://cs.opensource.google/go/go/+/master:src/net/http/httputil/reverseproxy.go;l=769

internal/ipc/hijack.go
joe0BAB: I think this can clean up a lot! Thanks for pointing it out! I got the tests green with this and doing a further bit of research this actually seems to be the way to do that upgrade cleanly (eg see https://github.com/coder/websocket/blob/91013c129147ee6375b65cf150fa151d2fcc0bce/accept.go#L150)

internal/ipc/hijack.go
```diff
@@ -0,0 +1,161 @@
...
+		return fmt.Errorf("clearing deadline: %w", err)
+	}
+	defer func() { _ = conn.SetReadDeadline(time.Time{}) }()
+	resp, err := http.ReadResponse(bufio.NewReader(conn), req)
+	if err != nil {
+		return err
+	}
+	if resp.Body != nil {
+		defer resp.Body.Close()
+	}
```
Benehiko: I don't think this is necessary. You can always defer the close after checking if error is nil. AFAIK, error being nil means Body is set, even if there is no data.

internal/ipc/hijack.go
```diff
@@ -0,0 +1,161 @@
...
+	if resp.StatusCode != http.StatusSwitchingProtocols {
+		var respBody []byte
+		if resp.Body != nil {
```
Benehiko: no need to check this, resp.Body cannot be empty if the http.response `err == nil`.

internal/ipc/hijack.go
```diff
@@ -0,0 +1,161 @@
...
+
+func writeAckWithTimeout(conn net.Conn, timeout time.Duration) error {
+	if err := conn.SetWriteDeadline(time.Now().Add(timeout)); err != nil {
+		return fmt.Errorf("clearing deadline: %w", err)
+	}
+	defer func() { _ = conn.SetWriteDeadline(time.Time{}) }()
+
+	// We tell the server that we have fully read the response and will now switch to serving gRPC
+	// by sending a single byte ack. This prevents the server from starting to send gRPC client
+	// traffic while we are still reading the previous HTTP response.
```
Benehiko: but we don't exclusively use this for GRPC? i think this comment is also a copy-paste from another source

internal/ipc/hijack.go
```diff
@@ -0,0 +1,161 @@
...
+
+func readAckWithTimeout(conn net.Conn, timeout time.Duration) error {
+	if err := conn.SetReadDeadline(time.Now().Add(timeout)); err != nil {
+		return fmt.Errorf("clearing deadline: %w", err)
+	}
+	defer func() { _ = conn.SetReadDeadline(time.Time{}) }()
+
+	// The client confirms it has fully read the response and is ready to serve gRPC by sending
+	// a single byte ack. This prevents race conditions where we start trying to connect gRPC clients
+	// before the response has been fully read, which can mix http and gRPC traffic.
```
Benehiko: here as well, the comment is a copy-paste. Not sure if it fully applies to us

Benehiko (approved):

PR #121, opened 2025-07-18
feat(sdk): engine supports externally launched plugins

### LINKED ISSUE #119: Engine: Integrate runtime for external plugins

joe0BAB:
Closes #119 .

Uses the hijacking layer to integrate the runtime for externally launched plugins with the secrets engine server.

Adds missing timeout handling on the hijacking logic.


engine/engine.go
```diff
@@ -184,14 +185,28 @@ func discoverPlugins(pluginPath string) ([]string, error) {
...
+			for conn := range acceptor.NextHijackedConn() {
+				launcher := Launcher(func() (runtime, error) {
+					return newExternalPlugin(conn, runtimeCfg{out: pluginCfgOut{engineName: cfg.name, engineVersion: cfg.version, requestTimeout: getPluginRequestTimeout()}})
+				})
+				if err := register(reg, launcher); err != nil {
+					logrus.Errorf("registering dynamic plugin: %v", err)
+				}
+			}
+		}()
+	}
```
joe0BAB: This is the main change. Everything else essentially is adjusting/adding tests and propagating things through.

internal/ipc/hijack.go
```diff
@@ -55,7 +59,18 @@ func Hijackify(conn net.Conn) (net.Conn, error) {
...
 
-	req, err := http.NewRequest("GET", "http://secrets-engine.localhost"+hijackPath, nil)
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	// This is a bit nasty as we overwrite any potential earlier set ReadDeadline
+	// and thereby erase it on the defer.
+	// However, passing in a connection here means we're going to own it.
+	// -> Overwriting any previous ReadDeadline is OK and any following usage of conn has to manage
+	// timeouts properly on its own anyway.
+	if dl, ok := ctx.Deadline(); ok {
```
joe0BAB: This popped up while writing tests: The was timeout management missing.

engine/engine.go
```diff
@@ -184,14 +185,28 @@ func discoverPlugins(pluginPath string) ([]string, error) {
...
 		w.WriteHeader(http.StatusOK)
 		_, _ = w.Write([]byte("ok"))
 	})
 	r := &resolver{reg: reg}
 	httpMux.Handle(resolverv1connect.NewResolverServiceHandler(&resolverService{r}))
+	if !cfg.dynamicPluginsDisabled {
+		acceptor := ipc.NewHijackAcceptor()
+		httpMux.Handle(acceptor.Handler())
+		go func() {
+			for conn := range acceptor.NextHijackedConn() {
```
Benehiko: so this means each plugin spun up would have to wait for ACK of the previous plugin, if no ACK it waits 2 seconds, so it could cascade to a long wait time until all plugins are registered. 

The for loop is okay, but we could also spin up a go routine per connection as well

Benehiko (approved):

engine/engine.go
joe0BAB: no, the ACK has already happened before and is parallelized by the server. However, you're right that the `register` is not parallelized yet and could lead to a bottle neck. Let me think about it.

internal/ipc/hijack.go
```diff
@@ -36,64 +36,77 @@ func Hijackify(conn net.Conn) error {
...
+		return nil, fmt.Errorf("unexpected status %d: %s", resp.StatusCode, string(respBody))
 	}
-	return nil
+	// If there is buffered content, wrap the connection.  We return an
+	// object that implements CloseWrite if the underlying connection
+	// implements it.
+	if _, ok := conn.(CloseWriter); ok {
+		return &hijackedConnCloseWriter{&hijackedConn{conn, br}}, nil
+	}
+	return &hijackedConn{conn, br}, nil
```
joe0BAB: @Benehiko this is the fix to replace the ACK with using net.Conn properly. I've changed from

```
if br.Buffered() > 0 {
		// If there is buffered content, wrap the connection.  We return an
		// object that implements CloseWrite if the underlying connection
		// implements it.
		if _, ok := c.(CloseWriter); ok {
			c = &hijackedConnCloseWriter{&hijackedConn{c, br}}
		} else {
			c = &hijackedConn{c, br}
		}
	} else {
		br.Reset(nil)
	}
```
To removing the ` br.Buffered() > 0` check and not doing that weird Reset thing but always just keep using the buffer.

Benehiko: But can't we always have a `CloseWriter` on `hijackedConn`? And then instead check inside `CloseWriter` if the underlying type is of `CloseWriter`?

PR #129, opened 2025-07-23
feat: configurable logging

### LINKED ISSUE #75: Setup Logging (plugin system)
TODO

joe0BAB:
This PR tests the waters with introducing a proper concept to logging. It mainly serves as validation if the suggested logging concept that is leaned on https://github.com/containerd/log is useful to our use cases. Fully refactoring to it is left to follow up PRs.


engine/engine.go
```diff
@@ -130,7 +130,7 @@ func newLauncher(cfg config, pluginFile string) (string, Launcher) {
 
 type Launcher func() (runtime, error)
 
-func register(reg registry, launch Launcher) error {
+func register(logger logging.Logger, reg registry, launch Launcher) error {
```
joe0BAB, line 134: I purposely decided against introducing a context here: This is a logic helper function around lifecycle management. So introducing a context here that effectively is yet another lifecycle thing could easily confuse things here even more. (this function will get more complex soon when I implement restart logic)

internal/logging/logging.go
```diff
@@ -0,0 +1,54 @@
...
+type loggerKey struct{}
+
+// WithLogger returns a new context with the provided logger.
+func WithLogger(ctx context.Context, logger Logger) context.Context {
+	return context.WithValue(ctx, loggerKey{}, logger)
+}
+
+// FromContext retrieves the current logger from the context. If no logger is
+// available, a new default logger gets returned.
+func FromContext(ctx context.Context) Logger {
```
joe0BAB: The AI didn't like `GetLogger` but kept suggesting/autocompleting `FromContext` so I went with it 🙈

Benehiko: We should error here when no logger is found instead of creating a new one.
Creating a new logger is too implicit that you won't realize you are using the wrong logger instead of fixing the value in the context

engine/engine.go
```diff
@@ -34,7 +33,7 @@ func newEngine(cfg config) (io.Closer, error) {
 	}
 
 	reg := &manager{}
-	startBuiltins(context.Background(), reg, cfg.plugins)
+	startBuiltins(logging.WithLogger(context.Background(), cfg.logger), reg, cfg.plugins)
```
Benehiko: we should avoid `context.Background()`. Instead we should opt for `context.WithoutCancel` https://pkg.go.dev/context#WithoutCancel and always reference a parent context. 

The start of the program can use `context.Background` but that's the only place - in future with OTEL we would want contexts to be related rather than disjointed.

joe0BAB: Should we bubble that up to the top even through the exported `New`, ie change it to `engine.New(context, name, ...)`?

joe0BAB: Or is `New` where within we'd create `context.Background`?

Benehiko: i think we should bubble it up as far as possible and let whoever is calling the secrets engine pass in their context.

internal/ipc/hijack.go
```diff
@@ -119,13 +122,15 @@ func (h *hijackHandler) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
...
 	conn, _, err := hj.Hijack()
 	if err != nil {
-		logrus.Errorf("hijack error: %v", err)
+		h.logger.Errorf("hijack error: %v", err)
 		return
 	}
 
+	// Note: Don't try to pass r.Context() in here.
+	// r.Context() gets cancelled when this function returns, but conn will live much longer which is misleading.
 	h.cb(conn)
```
Benehiko: you still can pass it, but use Context.WithoutCancel

Benehiko (approved):

PR #130, opened 2025-07-23
store: secret.Metadata/secret.SetMetadata and store.GetAllMetadata

Benehiko:
This patch adds a new `secret.Metadata()` and `secret.SetMetdata()`
function to the `secret` interface. This allows the underlying store
from storing non-sensitive metadata and retrieving only the metadata.

This patch also changes the behavior of the store.GetAll function so that it
always sets metadata, but does not set the actual secret from the
underlying keychain.

It's very costly to read each secret from the keychain since each value
needs to be decrypted. On macOS the user would be prompted for each value
decrypted.


store/keychain/keychain_darwin.go
```diff
@@ -143,6 +164,19 @@ func (k *keychainStore[T]) Save(_ context.Context, id store.ID, secret store.Sec
 	// it is a user-friendly name for the item, which is displayed in the keychain UI.
 	// https://developer.apple.com/documentation/security/ksecattrlabel
 	item.SetLabel(k.itemLabel(id))
+
+	metadata := make(map[string]string)
+	maps.Copy(metadata, secret.Metadata())
+	metadata["id"] = id.String()
+	metadata["service:group"] = k.serviceGroup
+	metadata["service:name"] = k.serviceName
```
joe0BAB, line 172: don't we need some kind of check here for overwriting any previous metadata? Not sure what's the best solution here. But it feels a bit we have certain reserved keys and if `secret.Metadata()` clashes with them, we return an error (a bit like python does it maybe with there `__something` thing, eg `__id`, maybe something like this could work and then documenting of course)

store/keychain/keychain_linux.go
```diff
@@ -318,6 +319,8 @@ func (k *keychainStore[T]) Save(_ context.Context, id store.ID, secret store.Sec
 	}
 
 	attributes := newItemAttributes(id.String(), k)
+	maps.Copy(attributes, secret.Metadata())
```
joe0BAB, line 322: Same issue as on mac: protect that we don't overwrite our internal attributes

store/keychain/keychain_test.go
```diff
@@ -110,17 +129,34 @@ func TestKeychain(t *testing.T) {
...
+			v.Password = ""
+
+			// the store sets some attributes internally, which we need to setup
+			// here on our expected map.
+			if v.Attributes == nil {
+				v.Attributes = make(map[string]string)
+			}
+			v.Attributes["id"] = k.String()
+			v.Attributes["service:group"] = "com.test.test"
+			v.Attributes["service:name"] = "test"
```
joe0BAB, line 156: This is a bit surprising. As a user, I'd not expect to see attributes that I haven't set myself. We should either document this behavior or update the implementation that our internal attributes get stripped / excluded in GetAll etc.

store/keychain/keychain_test.go
```diff
@@ -197,7 +233,7 @@ func TestKeychain(t *testing.T) {
 			require.NoError(t, kc.Delete(t.Context(), id))
 		})
 		require.NoError(t, kc.Save(t.Context(), id, &mustUnmarshalError{}))
-		_, err = kc.GetAll(t.Context())
+		_, err = kc.GetAllMetadata(t.Context())
 		require.ErrorContains(t, err, "i am failing on purpose")
 	})
```
joe0BAB, line 238: We're missing some tests about our internal attributes clashing with the user also setting attributes with same keys and how it get's handled.

store/keychain/keychain_darwin.go
joe0BAB: I guess alternatively, we could also have some kind of clash algorithm that on clash somehow encodes that in the key and reverse that later (like we'd apply that algorithm to user keys so they never clash)

store/keychain/keychain_windows.go
```diff
@@ -116,7 +120,26 @@ func findServiceCredentials[T store.Secret](k *keychainStore[T], credentials []*
...
+	return winAttrs
+}
+
+func mapFromWindowsAttributes(winAttrs []wincred.CredentialAttribute) map[string]string {
+	attributes := make(map[string]string, len(winAttrs))
+	for _, attr := range winAttrs {
+		attributes[attr.Keyword] = string(attr.Value)
+	}
+	return attributes
+}
```
joe0BAB, line 140: We easily could (and should) write a roundtrip unit test for this.

store/keychain/keychain_windows.go
```diff
@@ -161,24 +174,18 @@ func (k *keychainStore[T]) Save(_ context.Context, id store.ID, secret store.Sec
 		return err
 	}
 
+	attributes := make(map[string]string)
+	maps.Copy(attributes, secret.Metadata())
+
+	attributes["id"] = id.String()
+	attributes["service:group"] = k.serviceGroup
+	attributes["service:name"] = k.serviceName
```
joe0BAB, line 182: same overwrite issue as everywhere else

store/store.go
```diff
@@ -48,8 +54,12 @@ type Store interface {
...
 	// Get retrieves credentials from the store for a given ID.
 	Get(ctx context.Context, id ID) (Secret, error)
-	// GetAll retrieves all the credentials from the store.
-	GetAll(ctx context.Context) (map[ID]Secret, error)
+	// GetAllMetadata retrieves all the credentials from the store.
+	// Credentials retrieved will only call [Secret.SetMetadata] so that the
+	// underlying store does not get queried for each secret's sensitive data.
+	// This could be very taxing on the underlying store and cause a poor User
+	// Experience.
+	GetAllMetadata(ctx context.Context) (map[ID]Secret, error)
```
joe0BAB, line 62: Right now this has some undocumented behavior of returning internal attributes per secret as well.

joe0BAB (approved):

PR #131, opened 2025-07-23
tests: consolidate channel timeout/wait logic

### LINKED ISSUE #75: Setup Logging (plugin system)

joe0BAB:
This was becoming a bit too excessive with the same problem repeating everywhere and always re-hacking the same util function so pushed them into their own package (plus I'll need more of these in follow up PRs).

Also someone recently made me watch a talk about Go generics...


internal/testhelper/testing.go
```diff
@@ -0,0 +1,45 @@
...
+
+func WaitForWithExplicitTimeout[T any](ch <-chan T, timeout time.Duration) error {
+	select {
+	case val, ok := <-ch:
+		if !ok {
+			return nil
+		}
+		if err, isErr := any(val).(error); isErr {
+			return err
+		}
```
Benehiko: im not so sure about this one. Perhaps we should instead return the value T instead and ask the caller to manage that check?

internal/testhelper/testing.go
```diff
@@ -0,0 +1,45 @@
...
+func WaitForWithExplicitTimeoutV[T any](ch <-chan T, timeout time.Duration) (T, error) {
+	var zero T
+	select {
+	case val, ok := <-ch:
+		if !ok {
+			return zero, errors.New("channel closed")
+		}
+		if err, isErr := any(val).(error); isErr {
+			return val, err
+		}
```
Benehiko: same here

internal/testhelper/testing.go
joe0BAB: So T effectively only is `error` and `struct{}` and in case of `error`, we care about the value, otherwise not. I don't want to pollute my test code with putting the check everywhere as that would defeat the whole purpose of factoring this out for reducing boilerplate. The alternative would be to not have this generic function but two explicit ones (or not export the generic version but only the explicit versions).

internal/testhelper/testing.go
joe0BAB: right, here we indeed don't need this

internal/testhelper/testing.go
Benehiko: doing this check inside the generic function kind of defeats the purpose of generics and makes it non-obvious that an underlying return value of type error can be returned under the same value as timeout error, which seems iffy to me. 

I'd be fine with a function that calls the generic and does the check for you, that is explicit about it.

joe0BAB: That actually does not work due to the return type but I'll rewrite into two explicit functions then.

Benehiko (approved):

PR #133, opened 2025-07-23
store/keychain: manage internal attributes internally

Benehiko:
This patch keeps internal attributes stored on secrets from leaking to the consumer.

PR #135, opened 2025-07-24
keychain: metadata must not clash with internal

Benehiko:
Any metadata set on a secret inside the keychain must be prefixed to avoid clashes with internally used metadata.

Related: https://github.com/docker/secrets-engine/pull/108/ and https://github.com/docker/secrets-engine/pull/133


store/keychain/keychain_test.go
```diff
@@ -237,3 +237,100 @@ func TestKeychain(t *testing.T) {
 		require.ErrorContains(t, err, "i am failing on purpose")
 	})
 }
+
```
joe0BAB, line 240: Could we merge TestSafelySetMetadata and TestSafelyCleanMetadata so can have full roundtrip tests, ie, verify that after applying setMetadata and then cleanMetadata we get back something that is equal to the original?

Like always make sure that for any test the roundtrip at the end gives back the original data.

Benehiko: Didn't want to merge the two different tests, instead added a new test.

store/keychain/keychain_test.go
```diff
@@ -237,3 +237,134 @@ func TestKeychain(t *testing.T) {
...
+}
+
+func TestInternalMetadata(t *testing.T) {
+	kc := &keychainStore[*mocks.MockCredential]{
+		serviceGroup: "com.test.test",
+		serviceName:  "test",
+	}
+
+	t.Run("metadata can safely be set and cleaned afterwards", func(t *testing.T) {
+		attributes := map[string]string{
```
joe0BAB, line 345: could we do something like `original := maps.Copy(attributes)` here and use that down below instead of relying on manually retyping everything?

Benehiko: i prefer the manual typing, it makes it obvious what we expect

joe0BAB (approved):

PR #139, opened 2025-07-24
feat(sdk): retry logic loop

### LINKED ISSUE #87: Restart Engine Launched Plugins on Crash

joe0BAB:
Sets up retry loop logic.


engine/engine.go
```diff
@@ -132,26 +132,63 @@ func newLauncher(cfg config, pluginFile string) (string, Launcher) {
 
 type Launcher func() (runtime, error)
 
-func register(logger logging.Logger, reg registry, launch Launcher) error {
+func retryLoop(done <-chan struct{}, cfg config, reg registry, name string, launch Launcher) error {
+	counter := 0
+	var lastErr error
+	var errClosed <-chan error
+	for counter == 0 || cfg.retryLimit < 0 || counter <= cfg.retryLimit {
```
Benehiko: let's maybe just re-use backoff retry? in that way we can retry wait instead of just retry. The code would also be a lot cleaner here. https://github.com/cenkalti/backoff

joe0BAB: Good call! I'll rewrite things and go with that. Then we can get rid of all the tests as well.

engine/engine.go
```diff
@@ -132,26 +134,57 @@ func newLauncher(cfg config, pluginFile string) (string, Launcher) {
 
 type Launcher func() (runtime, error)
 
-func register(logger logging.Logger, reg registry, launch Launcher) error {
+func RetryLoop(ctx context.Context, cfg config, reg registry, name string, launch Launcher) error {
```
joe0BAB, line 137: Only temporary export to make the linter happy.

engine/engine.go
```diff
@@ -132,26 +134,57 @@ func newLauncher(cfg config, pluginFile string) (string, Launcher) {
...
+	cfg.logger.Printf("registering plugin '%s'...", name)
+
+	opts := []backoff.RetryOption{
```
Benehiko, line 142: you can also add retry delay here and max retry delay, e.g. retry and wait, retry and wait longer until maxRetryWait.

Benehiko (approved):

engine/engine.go
joe0BAB: Good callout, tweaked the defaults a bit to our use case

PR #141, opened 2025-07-25
refactor: secrets.ID and secrets.Pattern must always be valid

### LINKED ISSUE #147: secret.ID and secret.Pattern should always be valid
The `secrets.ID` and `secrets.Pattern` types are actually just string types, which means anyone can pass along an invalid `secrets.ID` or `secrets.Pattern`. Downstream functions would never be able to know if an ID is valid and any function returning the value could also return an invalid ID (even if it's by mistake). We should ensure that the value can reliably be created without needing to call `IsValid` all the time.

For example:

```go
func store(id secrets.ID) error {
    // can store id without requiring isValid here
}
```

To accomplish this, we can instead use an interface with an unexported struct.

```go
type id struct {
    val string
}

func NewID(s string) (ID, error) {
    if !valid(s) {
        return nil, errors.New("id not valid")
    }
    return &id{s}, nil
}
```

Benehiko:
closes https://github.com/docker/secrets-engine/issues/147

While using the `secrets.ID` and `secrets.Pattern` it became obvious that the caller would need take the responsibility of validating them before using the `Match` function. The root cause of this is that any string could be an ID/Pattern which means no validation is necessary to create one.

This patch changes that and instead of relying on the caller to validate, we enforce it instead on creation of the ID/Pattern type.


internal/secrets/pattern.go
```diff
@@ -2,39 +2,45 @@ package secrets
...
-	"fmt"
 )
 
 var ErrInvalidPattern = errors.New("invalid pattern")
 
 // Pattern can be used to match secret identifiers.
 // Valid patterns must follow the same validation rules as secret identifiers, with the exception
 // that '*' can be used to match a single component, and '**' can be used to match zero or more components.
-type Pattern string
+type Pattern struct {
```
joe0BAB: Could we keep this internal and instead export an interface only?

internal/secrets/secrets.go
```diff
@@ -13,7 +13,7 @@ var (
 )
 
 type Request struct {
-	ID ID `json:",omitzero"`
+	ID *ID `json:",omitzero"`
```
joe0BAB: Changing to reference like this smells. What about using an interface?

Benehiko:
Closing this as the PR was split up into multiple smaller ones.

PR #142, opened 2025-07-25
feat(sdk): retry supervisor layer

### LINKED ISSUE #87: Restart Engine Launched Plugins on Crash

joe0BAB:
This PR adds a new abstraction layer to introduce the retry logic. 

Before: Builtin plugins and discovered plugins would be launched directly

After: Builtin plugins and discovered plugins get wrapped with their launcher + retry logic and then later started by the retry supervisor layer. On shutdown, first the supervisor layer is stopped, then all plugins.

Closes #87 .


engine/engine_test.go
```diff
@@ -174,61 +174,177 @@ func Test_discoverPlugins(t *testing.T) {
...
 	})
 	t.Run("empty dir string", func(t *testing.T) {
-		plugins, err := discoverPlugins(testLogger(t), "")
+		plugins, err := scanPluginDir(testLogger(t), "")
 		assert.NoError(t, err)
 		assert.Empty(t, plugins)
 	})
 }
 
-func Test_startPlugins(t *testing.T) {
```
joe0BAB, line 196: Testing this separate no longer makes sense. It's now tested as part of the engine test.

engine/engine_test.go
```diff
@@ -174,61 +174,182 @@ func Test_discoverPlugins(t *testing.T) {
...
-	require.NoError(t, err)
-	t.Cleanup(func() { assert.NoError(t, e.Close()) })
-	c, err := client.New(client.WithSocketPath(socketPath))
-	require.NoError(t, err)
-	foo, err := c.GetSecret(t.Context(), secrets.Request{ID: "foo"})
-	assert.NoError(t, err)
-	assert.Equal(t, secrets.ID("foo"), foo.ID)
-	assert.Equal(t, "foo-value", string(foo.Value))
+	t.Parallel()
+	t.Run("happy path", func(t *testing.T) {
```
Benehiko: give more descriptive names here, what is the `happy path`?

engine/engine_test.go
```diff
@@ -174,61 +174,182 @@ func Test_discoverPlugins(t *testing.T) {
...
+		e, err := newEngine(testLoggerCtx(t), cfg)
+		require.NoError(t, err)
+		t.Cleanup(func() { assert.NoError(t, e.Close()) })
+		assert.EventuallyWithT(t, func(collect *assert.CollectT) {
+			assert.ElementsMatch(collect, e.Plugins(), []string{"plugin-bar-2-1"})
+		}, 2*time.Second, 100*time.Millisecond)
+		c, err := client.New(client.WithSocketPath(socketPath))
+		require.NoError(t, err)
+		bar, err := c.GetSecret(t.Context(), secrets.Request{ID: "bar"})
+		assert.NoError(t, err)
```
Benehiko: we should use `require` here on this line instead of assert. Was diagnosing nil pointers in the refactor I was doing, since the test continued after `GetSecret` error'ed which cause the following tests to never execute. 

Tests don't have recovery from nil pointers, they immediately exit and crash the rest of the tests.

engine/supervisor.go
```diff
@@ -0,0 +1,59 @@
...
+
+	for name, run := range initialProcesses {
+		downGroup.Add(1)
+
+		go func() {
+			err := run(ctxChild)
+			if err != nil && !errors.Is(err, context.Canceled) {
+				cfg.logger.Errorf("plugin '%s' stopped: %s", name, err)
+			}
+			downGroup.Done()
```
Benehiko: perhaps move the `downGroup.Done()` to the top of this function and use a `defer` instead. Although it makes no difference at this current moment, we don't want to accidentally introduce regressions.

engine/engine_test.go
```diff
@@ -174,61 +174,182 @@ func Test_discoverPlugins(t *testing.T) {
...
+		select {
+		case blockRunCh <- struct{}{}:
+		case <-time.After(5 * time.Second):
+			t.Fatal("timeout waiting for block run")
+		}
+		assert.EventuallyWithT(t, func(collect *assert.CollectT) {
+			assert.ElementsMatch(collect, e.Plugins(), []string{"my-builtin"})
+		}, 5*time.Second, 100*time.Millisecond) // Timeout needs to be larger than the initial retry interval
+		mySecret, err := c.GetSecret(t.Context(), secrets.Request{ID: "my-secret"})
+		assert.NoError(t, err)
```
Benehiko:
```suggestion
		require.NoError(t, err)
```

engine/engine_test.go
```diff
@@ -174,61 +174,182 @@ func Test_discoverPlugins(t *testing.T) {
...
+			logger:     testLogger(t),
+			maxTries:   1,
+		}
+		e, err := newEngine(testLoggerCtx(t), cfg)
+		require.NoError(t, err)
+		t.Cleanup(func() { assert.NoError(t, e.Close()) })
+		c, err := client.New(client.WithSocketPath(socketPath))
+		require.NoError(t, err)
+		foo, err := c.GetSecret(t.Context(), secrets.Request{ID: "foo"})
+		assert.NoError(t, err)
```
Benehiko:
```suggestion
		require.NoError(t, err)
```

engine/engine_test.go
joe0BAB: Fair I callout! I think I need to do this more consequential. There are many existing places where require isn't used but we rely on the value to not be nil later.

Benehiko (approved):

PR #145, opened 2025-07-28
tests: move dummy plugin code into testhelper package

### LINKED ISSUE #87: Restart Engine Launched Plugins on Crash

joe0BAB:
This PR moves the dummy plugin code into the `testhelper/dummy` directory and adds a bit more functionality to allow crashing plugins (needed in next PR).


engine/plugin_test.go
```diff
@@ -149,17 +150,10 @@ func Test_newPlugin(t *testing.T) {
...
 				assert.ErrorContains(t, err, errGetSecret)
 				assert.NoError(t, p.Close())
 				r, err := parseOutput()
-				// TODO: investigate
-				// This keeps randomly failing on CI with:
-				// failed to unmarshal '': unexpected end of JSON input
-				//
-				// It means the plugin hasn't been shutdown yet though we believe it should have
-				// or the plugin somehow didn't manage to send stuff to STDOUT.
-				// Either way, when we try to parse what we believe is STDOUT, it's empty.
```
joe0BAB, line 162: already fixed https://github.com/docker/secrets-engine/pull/134

internal/testhelper/dummy/dummy.go
```diff
@@ -153,6 +159,9 @@ type dummyPluginResult struct {
 func (d *dummyPlugin) GetSecret(_ context.Context, request secrets.Request) (secrets.Envelope, error) {
 	d.m.Lock()
 	defer d.m.Unlock()
+	if d.cfg.CrashBehaviour != nil && len(d.result.GetSecret)+1 >= d.cfg.OnNthSecretRequest {
+		os.Exit(d.cfg.ExitCode)
+	}
```
joe0BAB, line 164: I need this later for testing the retry logic. Also this is the second main change, that the dummy plugin can be configured to crash on GetSecret requests.

Benehiko (approved):

PR #148, opened 2025-07-28
fix(sdk): external plugin runtime returns an error when the binary unexpectedly stops

### LINKED ISSUE #140: Bug: When a plugin unexpectedly stops but with exit code 0, it's not handled as an error

joe0BAB:
Closes #140 : When a external plugin binary unexpectedly exists with code 0, that still should be handled as an error.


engine/plugin.go
```diff
@@ -127,30 +127,33 @@ func newLaunchedPlugin(logger logging.Logger, cmd *exec.Cmd, v runtimeCfg) (runt
...
+	r, err := setup(logger, rwc, ipc.NewServerIPC, setIpcClosed, v, ipc.WithShutdownTimeout(getPluginShutdownTimeout()))
 	if err != nil {
 		rwc.Close()
 		cmdWrapper.Close()
 		return nil, err
 	}
 
+	errUnexpectedShutdown := make(chan error, 1)
+	shutdownRequested, setShutdownRequested := closeOnce()
+	shutdownEnded, setShutdownEnded := closeOnce()
```
joe0BAB: I couldn't find any better way to solve this. There are a bunch of synchronization things that need to be taken care of with their special needs. I wished there were a more holistic approach to this.

Benehiko (approved):

PR #150, opened 2025-07-29
fix(ipc): assign client/server roles based on the context

### LINKED ISSUE #149: Bug: yamux header issue

```
```

joe0BAB:
Another attempt to address #149 : I noticed that for externally launched plugins, the roles of client and server actually are the other way around. Unfortunately 5fa27a9c03e9785548f5abdb3607334e73776bf7 went to fare. So this PR corrects it and only swaps server/client yamux roles for externally launched plugins but for engine launched plugins the IPC server remains the secrets engine.


Benehiko (approved):
Just reword the PR body, I think it doesn't properly convey what is done here.

PR #152, opened 2025-07-29
fix(engine): correctly wait for runtimes to finish

### LINKED ISSUE #87: Restart Engine Launched Plugins on Crash

joe0BAB:
Initially, we only had to wait for the IPC to shutdown. However, for engine launched plugins, we also need for the process to terminate. We missed adding this. This PR fixes that.


engine/engine.go
```diff
@@ -193,9 +197,7 @@ func register(logger logging.Logger, reg registry, launch launcher) (<-chan erro
 	}
 	errClosed := make(chan error, 1)
 	go func() {
-		if run.Closed() != nil {
-			<-run.Closed()
-		}
```
joe0BAB, line 198: This actually was a bug and could hang forever.

engine/engine_test.go
```diff
@@ -144,26 +146,14 @@ func Test_Register(t *testing.T) {
...
-		errCh, err := register(testLogger(t), reg, l)
+		errCh, err := register(testLoggerCtx(t), reg, l)
 		assert.NoError(t, err)
 		assert.Equal(t, 0, reg.removeCalled)
 		close(r.closed)
 		<-reg.removed
 		assert.Equal(t, 1, reg.removeCalled)
 		assert.NoError(t, testhelper.WaitForErrorWithTimeout(errCh))
 	})
-	t.Run("runtime gets unregistered when runtime closed channel is nil", func(t *testing.T) {
```
joe0BAB, line 155: no longer relevant

engine/plugin.go
```diff
@@ -163,7 +163,16 @@ func newLaunchedPlugin(logger logging.Logger, cmd *exec.Cmd, v runtimeCfg) (runt
...
 		resolverClient: resolverv1connect.NewResolverServiceClient(r.client, "http://unix"),
 		close:          finalCloseEverything,
-		closed:         closed,
+		wait: func(ctx context.Context) {
+			select {
+			case <-closed:
+			case <-ctx.Done():
+			}
+			select {
+			case <-cmdWrapper.Closed():
```
joe0BAB, line 172: this was missing

joe0BAB: This is the main change of the PR.

joe0BAB:
I'll address this differently

PR #153, opened 2025-07-29
refactor(engine): explicitly copy registration data to enforce validation

joe0BAB:
Preparation work for #141 .


engine/setup.go
```diff
@@ -79,15 +80,16 @@ func setup(logger logging.Logger, conn io.ReadWriteCloser, setup ipc.Setup, cb f
...
 	if p.name != "" && in.name != p.name {
-		return nil, errors.New("plugin name cannot be changed when launched by engine")
+		return nil, nil, errors.New("plugin name cannot be changed when launched by engine")
 	}
 	if p.name == "" && in.name == "" {
-		return nil, errors.New("plugin name is required when not launched by engine")
+		return nil, nil, errors.New("plugin name is required when not launched by engine")
 	}
-	return &p.out, nil
+	return &pluginCfgIn{in.name, in.version, pattern}, &p.out, nil
```
joe0BAB, line 94: This in combination with `secrets.ParsePattern` at the function top is the main change.

engine/registration.go
```diff
@@ -85,16 +91,16 @@ func newRegistrationLogic(validator pluginCfgInValidator, result chan registrati
...
-func (l *registrationLogic) register(_ context.Context, cfg pluginCfgIn) (*pluginCfgOut, error) {
+func (l *registrationLogic) register(_ context.Context, cfg pluginCfgInUnvalidated) (*pluginCfgOut, error) {
 	l.m.Lock()
 	defer l.m.Unlock()
 	if l.done {
 		return nil, errors.New("cannot rerun registration")
 	}
 	l.done = true
-	out, err := l.validator.Validate(cfg)
+	inValid, out, err := l.validator.Validate(cfg)
```
Benehiko: we should use a better name, this reads "invalid" or "in...valid" xD

engine/setup.go
```diff
@@ -50,11 +51,11 @@ func setup(logger logging.Logger, conn io.ReadWriteCloser, setup ipc.Setup, cb f
 	var out pluginCfgIn
 	select {
 	case r := <-chRegistrationResult:
-		if r.err != nil {
+		if r.err != nil || r.cfg == nil {
 			i.Close()
 			return nil, fmt.Errorf("failed to register plugin: %w", r.err)
 		}
-		out = r.cfg
+		out = *r.cfg
```
Benehiko, line 58: could `out pluginCfgIn` not also be a pointer?

joe0BAB: Right now that would lead to dragging around a pointer everywhere. Let me do a separate cleanup PR for to address this.

Benehiko (approved):

PR #154, opened 2025-07-29
fix(ipc): server side hijacking logic

### LINKED ISSUE #149: Bug: yamux header issue

joe0BAB:
There were two issues with the current implementation:
* `net.Conn` is only guaranteed to be valid while the request handler hasn't returned. Ie., we need to make it wait for the code using it to finish before returning the request handler. Ownership of `net.Conn` can't be transferred out of the request handler as previously assumed.
* The response header needs to be written after the hijack.

Closes #149 .

Related issues:
* https://github.com/golang/go/issues/27408
* https://github.com/golang/go/issues/32314


internal/ipc/hijack.go
```diff
@@ -59,9 +45,13 @@ func hijackRequest(conn net.Conn, timeout time.Duration) (net.Conn, error) {
 		return nil, fmt.Errorf("making hijack request: %s", err)
 	}
 
-	if err := conn.SetReadDeadline(time.Now().Add(timeout)); err != nil {
+	if err := conn.SetReadDeadline(time.Now().Add(4 * time.Second)); err != nil {
+		return nil, fmt.Errorf("clearing deadline: %w", err)
+	}
+	if err := conn.SetWriteDeadline(time.Now().Add(4 * time.Second)); err != nil {
```
Benehiko: aren't these just durations? do we need `time.Now()`?

internal/ipc/hijack.go
```diff
@@ -126,6 +158,25 @@ func (h *hijackHandler) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
...
+	if err := conn.SetReadDeadline(time.Now().Add(4 * time.Second)); err != nil {
+		h.logger.Errorf("SetReadDeadline error: %v", err)
+		conn.Close()
+		return
+	}
+	if err := conn.SetWriteDeadline(time.Now().Add(4 * time.Second)); err != nil {
+		h.logger.Errorf("SetWriteDeadline error: %v", err)
+		conn.Close()
+		return
+	}
```
Benehiko: same here

internal/ipc/hijack.go
```diff
@@ -104,33 +105,46 @@ type CloseWriter interface {
...
+	// Flush the options to make sure the client sets the raw mode
+	brw.Flush()
+
+	resp := &http.Response{
+		StatusCode: http.StatusSwitchingProtocols,
+		Proto:      r.Proto,
+		Header:     w.Header(),
+	}
```
Benehiko: image

internal/ipc/hijack.go
```diff
@@ -104,33 +105,57 @@ type CloseWriter interface {
...
+			Header:     w.Header(),
+		}
+		if err := resp.Write(conn); err != nil {
+			h.logger.Errorf("writing response: %v", err)
+			conn.Close()
+			return
+		}
+
+		h.cb(r.Context(), conn)
+		close(done)
```
Benehiko: this should be a defer I think

internal/ipc/hijack.go
```diff
@@ -104,33 +105,57 @@ type CloseWriter interface {
...
+		brw.Flush()
+
+		resp := &http.Response{
+			StatusCode: http.StatusSwitchingProtocols,
+			Proto:      r.Proto,
+			Header:     w.Header(),
+		}
+		if err := resp.Write(conn); err != nil {
+			h.logger.Errorf("writing response: %v", err)
+			conn.Close()
```
Benehiko: you don't need conn.Close() here if we are already closing when we return the `ServeHTTP` func

internal/ipc/hijack.go
```diff
@@ -104,33 +105,57 @@ type CloseWriter interface {
...
 
-	// Note: Be careful when trying to pass r.Context() in here.
-	// r.Context() gets cancelled when this function returns, but conn will live much longer which is misleading.
-	h.cb(conn)
+	defer conn.Close()
+
+	done := make(chan struct{})
+	go func() {
+		// Flush the options to make sure the client sets the raw mode
+		brw.Flush()
```
Benehiko: are we sure about this flush now? we aren't doing anything with `brw` after this

internal/ipc/hijack.go
```diff
@@ -104,33 +105,57 @@ type CloseWriter interface {
...
+
+		resp := &http.Response{
+			StatusCode: http.StatusSwitchingProtocols,
+			Proto:      r.Proto,
+			Header:     w.Header(),
```
Benehiko, line 140: I don't think we are setting the `Connection: Upgrade` and `Upgrade: tcp` headers anymore.

internal/ipc/hijack.go
joe0BAB: indeed, it seems this should work: https://github.com/golang/go/blob/880ca333d708b957325b6ef4798699372d4c090e/src/net/http/httputil/reverseproxy.go#L797-L804

Benehiko (approved):

Issue #156, opened 2025-07-30
Inconsistent JSON tags and validation issues


Benehiko:
There are some structs that we are sending over the wire which have inconsistent JSON tags, such as [Envelope](https://github.com/docker/secrets-engine/blob/eff1218bf7cc068d76c4fd31719c1875270b1147/internal/secrets/secrets.go#L25-L33), [Request](https://github.com/docker/secrets-engine/blob/eff1218bf7cc068d76c4fd31719c1875270b1147/internal/secrets/secrets.go#L15-L23) and [PluginCfg](https://github.com/docker/secrets-engine/blob/eff1218bf7cc068d76c4fd31719c1875270b1147/internal/testhelper/dummy/dummy.go#L190-L197). 

We should be careful when Marshaling/Unmarshaling such structs as the data are not properly being handled, especially with validation regarding `secrets.ID` and `secrets.Pattern`.

Benehiko:
Disabled JSON marshal on `Envelope` and `Request` https://github.com/docker/secrets-engine/pull/182
Removed JSON marshal on plugin version type https://github.com/docker/secrets-engine/pull/168
Refactored `PluginCfg` so that it does not need custom JSON marshalling https://github.com/docker/secrets-engine/pull/171

PR #157, opened 2025-07-30
Implement custom JSON handling for various internal types

Benehiko:
closes #156 

This patch implements custom JSON handling for Envelope, PluginCfg,
Request, ID interface and Pattern interface.

There were also inconsistencies in the JSON produced by our types. Some
types would not set any JSON tag, leaving it up to the default behavior
of `encoding/json`. In some cases we would need to handling custom types
which are always exposed as an interface to the outside.

This patch also adds validation to the Marshal/Unmarshal of our types,
so that we avoid future problems where an ID could be invalid, or an
Envelope produces unexpected data over the wire.


Benehiko:
Not needed anymore.

PR #158, opened 2025-07-30
refactor: enforce validation of plugin data

joe0BAB:
Follow up to #153: Enforce strict validation of plugin data by design (plugin data is the engine internal meta data associated with a plugin).

I'll push that in a follow up PR further down to be used by the public interfaces of engine/plugin as well.


internal/api/defaults_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_NewPluginData(t *testing.T) {
+	t.Parallel()
+	t.Run("name is required", func(t *testing.T) {
+		_, err := NewPluginData(PluginDataUnvalidated{Version: "foo", Pattern: "*"})
+		assert.ErrorContains(t, err, "name is required")
```
Benehiko, line 14: I think ErrorContains will panic if err is nil. We should probably do a `require.Error(t, err)` before this assert.

joe0BAB: Just tested, you get a `Error:      	An error is expected but got nil.`. So should be fine.

Benehiko (approved):

PR #159, opened 2025-07-30
feat(plugin): add parameter to configure logger

### LINKED ISSUE #75: Setup Logging (plugin system)

joe0BAB:
Changes the plugin SDK part so a logger can be configured (#75 ). Additionally, as this comes with a change of the config data type this PR also makes sure the new checked version type introduced in #161 gets propagated through properly.


engine/builtin.go
```diff
@@ -10,28 +10,48 @@ import (
...
-func newInternalRuntime(ctx context.Context, name string, p Plugin) (runtime, error) {
-	logger, err := logging.FromContext(ctx)
-	if err != nil {
+type internalPluginData struct {
+	name    string
+	pattern secrets.Pattern
+	version api.Version
+}
+
+// TODO: This entire thing smells. Try to consolidate.
```
Benehiko, line 29: still todo?

Benehiko (approved):

engine/builtin.go
joe0BAB: yes, didn't want to extend the scope of this refactoring too much

PR #161, opened 2025-07-31
feat: version type

### LINKED ISSUE #75: Setup Logging (plugin system)

joe0BAB:
Wrapper type around version field to assure any "version" we pass around is validated according to how we define a valid version.

For now, any semver string without `v` prefix is a valid version.


internal/api/version.go
```diff
@@ -0,0 +1,85 @@
...
+func valid(s string) error {
+	if len(s) > 0 && s[0] == 'v' {
+		return ErrVPrefix
+	}
+	if s == "" {
+		return ErrEmptyVersion
+	}
+	if !semver.IsValid("v" + s) {
+		return fmt.Errorf("invalid version: %s", s)
+	}
```
Benehiko, line 83: bit confused by this, so we add `v` but we reject plugins from adding `v` prefix?

Benehiko (approved):

internal/api/version.go
joe0BAB: yeah, that's debatable. It feels redundant to me being forced to prefix everything with `v`. But also maybe we just want to not surprise people and just plain use `semver.IsValid`

Benehiko (changes requested):
Let's move the hijack godocs into its own PR with a nice commit description

Benehiko (approved):

PR #163, opened 2025-07-31
docs: describe the details and pitfalls around the hijacking logic

### LINKED ISSUE #149: Bug: yamux header issue

joe0BAB:
For implementation, see:

ad920bde8e7913b681c80b238b956809dcc701e2
92a4628ccb9c3830118d8ce91397b109e9dc73ba

Adds missing documentation.


internal/ipc/design.md
```diff
@@ -37,6 +37,24 @@ flowchart TD
 
 ---
 
+### 2025-07-31 Doing Hijacking Logic (the right way)
+
+There was an ongoing flakiness issue that started to plague CI since ~2 weeks and that had been hard to pin down but now is finally solved.
+A core issue to this is that hijacking in `net/http.Server` is not very well documented and 
+that there's a bunch of implicit behaviour that can easily be missed and lead to an implementation that seems to work except that it does not in certain percentage of case, and you get reliability issues on the long run.
```
Benehiko:
```suggestion
There was a bug in the engine's ipc handler which caused intermittent plugin registration failures. The CI became extremely flaky which indicated that there was an underlying problem. 

The core issue is that hijacking in `net/http.Server` would intermittently drop the connection which the plugin establishes. The cause was a mixture of undocumented lifecycle behavior of the underlying `net.Conn` that was hijacked from connection in `net/http.Server` and a missing first byte from the payload sent by the client (possibly).
```

internal/ipc/design.md
```diff
@@ -37,6 +37,24 @@ flowchart TD
...
+
+The main takeaways are:
+- The hijacked `net.Conn` is only guaranteed to be valid while the request handler hasn't returned. I.e., we need to make it wait for the code using it to finish before returning the request handler. Ownership of net.Conn can't be transferred out of the request handler as previously assumed.
+- The response header needs to be written **after** the hijack.
+
+Also studying [the standard lib reverseproxy](https://github.com/golang/go/blob/880ca333d708b957325b6ef4798699372d4c090e/src/net/http/httputil/reverseproxy.go#L768-L804) helps a lot to understand how to implement hijacking on the server correctly.
```
Benehiko:
```suggestion
Also studying [the standard lib reverseproxy](https://github.com/golang/go/blob/880ca333d708b957325b6ef4798699372d4c090e/src/net/http/httputil/reverseproxy.go#L768-L804) helped a lot to understand how to implement hijacking on the server correctly.
```

Benehiko: We should add that we studied the dagger code base and their hack as well as Moby

Benehiko: linking to their code would also be good.

internal/ipc/hijack.go
```diff
@@ -110,7 +110,17 @@ type hijackHandler struct {
 	logger     logging.Logger
 }
 
+// ServeHTTP hijacks the underlying connection of a http request.
+// That connection then can be reused to send any data, i.e., it gives a universal duplex connection to the client.
+// Important notes to keep this stable:
```
Benehiko:
```suggestion
// The underlying connection can then be reused to send any data, i.e., it gives a universal duplex connection to the client.
// Important notes to keep this stable:
```

internal/ipc/hijack.go
```diff
@@ -110,7 +110,17 @@ type hijackHandler struct {
...
+//   - The hijacked `net.Conn` is only guaranteed to be valid while the request handler hasn't returned.
+//     I.e., we need to make it wait for the code using it to finish before returning the request handler.
+//     Ownership of net.Conn can't be transferred out of the request handler as previously assumed.
```
Benehiko:
```suggestion
//     I.e., we need to make it wait for the code using it to finish before returning the request handler.
//     Ownership of net.Conn can't be transferred out of the request handler.
```

internal/ipc/hijack.go
```diff
@@ -110,7 +110,17 @@ type hijackHandler struct {
...
+//   - The response header needs to be written **after** the hijack.
+//
+// See https://github.com/golang/go/issues/27408 and https://github.com/golang/go/issues/32314 for more details.
```
Benehiko, line 121: please add a link to go's reverse proxy and moby

internal/ipc/hijack.go
```diff
@@ -150,6 +160,8 @@ func (h *hijackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		// Important: The server still owns net.Conn! When this handler returns,
+		// conn gets closed and there's nothing we can do about as net/http.Server will do it anyway later.
```
Benehiko:
```suggestion
		// conn gets closed
```

Benehiko (approved):
LGTM

Issue #162, opened 2025-07-31
Crashed launched plugins are difficult to debug


Benehiko:
It would be nice to have a way to get the `stdout` and `stderr` of a launched plugin. At the moment we just have the exit code, which isn't very useful to debug with. 

To get any sort of indication of what failed, we need to manually short circuit the test plugins in `dummy.go`.

```diff
func PluginCommand(t *testing.T, cfg PluginCfg) (*exec.Cmd, func() (*PluginResult, error)) {
	t.Helper()
	cfgStr, err := cfg.toString()
	assert.NoError(t, err)
	cmd := exec.Command(os.Args[0]) // run the test binary again
	var stdoutBuf, stderrBuf bytes.Buffer
-       cmd.Stdout = stdoutBuf
-       cmd.Stderr = stderrBuf
+       cmd.Stdout = os.Stdout
+       cmd.Stderr = os.Stdout
	cmd.Env = append(os.Environ(),
		"RUN_AS_DUMMY_PLUGIN=1",
		dummyPluginCfgEnv+"="+cfgStr,
	)
```

It would be best to get the output of the buffer returned to the caller so that the caller can log/comparison check the output.

PR #164, opened 2025-07-31
engine: log cmdWatchWrapper underlying error

Benehiko:


engine/exec.go
```diff
@@ -79,6 +80,10 @@ func launchCmdWatched(logger logging.Logger, name string, p proc, timeout time.D
 		if err != nil {
 			err = fmt.Errorf("plugin %s crashed: %w", name, err)
 		}
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			logger.Errorf("launchCmdWatched: %s", string(exitErr.String()))
```
joe0BAB, line 85: we already catch that here don't we? https://github.com/docker/secrets-engine/blob/b581ed64b6e77b0f53ee5d19f3b252cc3ac76d08/engine/plugin_test.go#L202

Benehiko: oof, yeah that was testing code

Benehiko:
I guess we should do a follow-up on this at some point. Debugging plugins in our tests were a pain, especially since we noop the output.

Benehiko:
Issue exists here https://github.com/docker/secrets-engine/issues/162

PR #165, opened 2025-08-04
secrets: identifier and pattern must always be valid

Benehiko:
Relates to: #141 

This is an initial PR to split the large refactor in #141. A follow-up PR will start introducing the usage of IDNew and PatternNew.

It used to be possible to easily pass around invalid ID and Patterns due to their underlying type alias being string. This would give the burden of creating an ID / Pattern correctly on the caller. This PR attempts to make it more difficult to create an invalid ID / Pattern. Packages relying on this type would thus be more assured of its validity without doing unnecessary validation checks.


internal/secrets/identifiers_test.go
```diff
@@ -81,3 +81,93 @@ func TestMatch(t *testing.T) {
...
+	tests := []struct {
+		name     string
+		input    string
+		expected error
+	}{
+		{"valid name with dash", "my-secretA9", nil},
+		{"valid name with dot", "my.secretA9", nil},
+		{"valid name with slash", "my/secret", nil},
+		{"valid name with underscore", "my_secretA9", nil},
+		{"invalid name with trailing slash", "my/secret/", ErrInvalidID{"my/secret/"}},
```
joe0BAB, line 95: nice, I like this!

internal/secrets/identifiers_test.go
```diff
@@ -81,3 +81,93 @@ func TestMatch(t *testing.T) {
 		})
 	}
 }
+
+func TestParseIDNew(t *testing.T) {
+	tests := []struct {
```
joe0BAB, line 86: `t.Parallel()`? (effectively probably doesn't matter that much)

internal/secrets/identifiers_test.go
```diff
@@ -81,3 +81,93 @@ func TestMatch(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := ParseIDNew(tc.input)
+			assert.ErrorIs(t, err, tc.expected)
+		})
+	}
+}
+
+func TestMatchNew(t *testing.T) {
+	tests := []struct {
```
joe0BAB, line 111: `t.Parallel()`?

joe0BAB (approved):

internal/secrets/identifiers_test.go
Benehiko: will add those in another commit

PR #167, opened 2025-08-04
store(refactor): use the new secrets ID and Pattern types

Benehiko:
This patch refactors the store internals to use the new secrets ID and Pattern types. This allows the store to assume the values are always valid.


joe0BAB (approved):
so clean now! :smiling

PR #174, opened 2025-08-04
vendor: remove logrus as a direct dependency

Benehiko:
Logrus is not used anymore, somehow this was not caught by the CI. Locally I caught this in a few PRs with `make lint`.


joe0BAB (approved):
I believe the issue was that two PRs got merged that removed logrus. For each PR though, it wasn't removed entirely, only after both PRs were merged. We don't run CI again after merge -> it slipped through

PR #181, opened 2025-08-04
refactor: use checked types

joe0BAB:
Use the checked types on the config and take advantage of the new types to simplify the internal runtime interface.


engine/plugin.go
```diff
@@ -73,7 +73,7 @@ type runtime interface {
 
 	io.Closer
 
-	Data() api.PluginData
+	metadata
```
joe0BAB, line 76: main change

engine/adaptation.go
```diff
@@ -35,28 +35,29 @@ type Config struct {
 	// Version of the plugin in semver format.
 	Version api.Version
 	// Pattern to control which IDs should match this plugin. Set to `**` to match any ID.
-	Pattern secrets.Pattern
+	Pattern secrets.PatternNew
```
joe0BAB, line 38: main change

internal/api/name.go
```diff
@@ -0,0 +1,48 @@
+package api
+
+import (
+	"errors"
+	"fmt"
+)
+
+var ErrEmptyName = errors.New("empty name")
+
+type Name interface {
```
joe0BAB, line 10: introducing this allows more consistency

Benehiko (approved):

PR #182, opened 2025-08-04
secrets: disable JSON marshal on Envelope and Request

Benehiko:
This patch disables JSON marshal for secrets.Envelope and secrets.Request. It introduces a custom marshaler on these types which will panic. This is a bit strict, but we don't want to use JSON on these types unless absolutely necessary.


internal/secrets/secrets.go
```diff
@@ -13,24 +14,36 @@ var (
...
 type Envelope struct {
-	ID         ID
-	Value      []byte    `json:",omitzero"`
-	Provider   string    `json:",omitzero"`
-	Version    string    `json:",omitzero"`
-	Error      string    `json:",omitzero"`
-	CreatedAt  time.Time `json:",omitzero"`
-	ResolvedAt time.Time `json:",omitzero"`
-	ExpiresAt  time.Time `json:",omitzero"`
+	ID         ID        `json:"-"`
```
joe0BAB, line 33: wouldn't we want to remove the json annotation entirely then?

Benehiko: no, let's leave it like this. This indicates to future maintainers that we completely omit the field anyway. Removing the tag (and custom Marshaller) would have the exported field marshalled automatically.

joe0BAB (approved):

PR #186, opened 2025-08-04
refactor: remove old secrets.ID and secrets.Pattern

Benehiko:
This PR removes the old `secrets.ID` and `secrets.Pattern` type and replaces it with the new types.


joe0BAB (approved):
🎉

PR #191, opened 2025-08-05
fix: version strings must be prefixed with 'v'

joe0BAB:
Version strings should be prefixed with `v`. This makes them easier to parse/read as a human and identify as version strings vs just a number string that hasn't been parsed to int (eg `"v1"` vs `"1"`). `golang.org/x/mod/semver` does it like this by default and I learned the hard way why this is a good idea.


internal/api/version.go
```diff
@@ -46,13 +46,13 @@ func MustNewVersion(s string) Version {
...
 func valid(s string) error {
-	if len(s) > 0 && s[0] == 'v' {
+	if len(s) > 0 && s[0] != 'v' {
 		return ErrVPrefix
 	}
 	if s == "" {
 		return ErrEmptyVersion
 	}
-	if !semver.IsValid("v" + s) {
+	if !semver.IsValid(s) {
```
joe0BAB, line 55: main change

Benehiko (approved):

PR #193, opened 2025-08-05
secrets: pattern can be split into parts

Benehiko:
A pattern might need to be split up into "parts". In this case each part should hold its "part type" which indicates if it's an exact match string or an any / any recursive.

Relates to https://github.com/docker/secrets-engine/pull/108


Benehiko:
not necessary anymore

PR #201, opened 2025-08-08
feat(my secret): set secret command

### LINKED ISSUE #188: Setup: new binary, docker cli plugin boilerplate, release pipeline, ...
Initial plumbing work to get a new binary we can then iterate on to build the mysecret CLI.

### LINKED ISSUE #200: Basic operations: set/get/rm/list

joe0BAB:
Setups up cross compilation for the mysecret plugin binary and adds the `docker mysecret set` command.

Completes #188 and starts #200 .


mysecret/go.mod
```diff
@@ -0,0 +1,76 @@
+module github.com/docker/secrets-engine/mysecret
+
+go 1.24.3
+
+replace github.com/docker/secrets-engine => ../
```
Benehiko, line 5: let's not use replace directives anymore. it obscures problems with go get.

mysecret/go.mod
```diff
@@ -0,0 +1,76 @@
...
+
+replace github.com/docker/secrets-engine/store => ../store
+
+require (
+	github.com/docker/cli v28.3.3+incompatible
+	github.com/docker/secrets-engine v0.0.0-00010101000000-000000000000
+	github.com/docker/secrets-engine/store v0.0.0-00010101000000-000000000000
```
Benehiko: use the latest version release (v0.0.7)

mysecret/commands/set.go
```diff
@@ -0,0 +1,119 @@
...
+	}()
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case err := <-errs:
+		return nil, err
+	case line := <-lines:
+		return line, nil
+	}
```
Benehiko: this function doesn't convince me. It might be racy when the defer closes errs and lines triggering the select incorrectly, even if there is data in lines. can't we instead just use a scanner? perhaps something like this https://github.com/docker/cli-experiments/pull/58/files#diff-b5ea557a04ebc29f20b69f87cd6d5c60635b3021c6847ae75f1fea92e42e5dddR69

mysecret/commands/set.go
```diff
@@ -0,0 +1,119 @@
...
+	}, nil
+}
+
+type secret struct {
+	id  string
+	val string
+}
+
+func parseArg(arg string) (*secret, error) {
+	parts := strings.Split(arg, "=")
```
Benehiko, line 84: wouldn't we want to check if the ID conforms to the secrets.ID here?

joe0BAB: validation is done later when we do `secrets.ParseID`

If we do it here, we'd need to do it twice as there are two ways we can get an ID. 

It's parsing (different depending if we read from STDIN or not) -> validating (same for both)

mysecret/go.mod
joe0BAB: I can't, due to `go work sync` we need to change them all at once which is done in #199

mysecret/go.mod
joe0BAB: I can't, due to `go work sync` we need to change them all at once which is done in #199

mysecret/commands/set.go
joe0BAB: Indeed. I'm not sure about scanner either as we want to read everything, but we can use `io.Copy` I think. Also let's not get hung up too much on this in this PR as I'll do a follow up PR to add tests for this anyway (and that should reveal all the issues around this).

mysecret/go.mod
Benehiko: okay, then at least specify the tag for the secrets-engine and secrets-engine/store modules inside go.mod so that we can pin it to a version instead of a pseudo version

joe0BAB: right, this seems to work, changed it

mysecret/commands/set.go
```diff
@@ -0,0 +1,109 @@
...
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case err := <-done:
+		if err != nil {
+			return nil, err
+		}
+		return buf.Bytes(), nil
+	}
```
Benehiko, line 108: what's wrong with scanner? I'd rather use scanner here since io.Copy won't EOF on pressing Carriage return. The solution I provided here is solid https://github.com/docker/cli-experiments/pull/58/files#diff-b5ea557a04ebc29f20b69f87cd6d5c60635b3021c6847ae75f1fea92e42e5dddR69

joe0BAB: How do you press carriage return on something that pipes to STDIN? Any chance you're mixing this with reading user input? This is about reading piped input.

Benehiko: i thought this function was for both cases

joe0BAB: But then it's different behavior as we'd only scan until `\n` and not copy everything. Those are two different things. If we'd abort on `\n` here, then let's say you did `cat foo.txt > mysecret set foo` we'd only read the first line of `foo.txt`.

Benehiko (approved):

mysecret/commands/set.go
Benehiko: hmm i wonder about that. piping a file would probably also have the value in 1 line? e.g. 

foo.txt
```
key:value
```

mysecret/commands/set.go
joe0BAB: So I followed up on this and wrote a unit tests here: https://github.com/docker/secrets-engine/pull/228

The directly linked solution doesn't work as it only scans one line. Ie., we'd need to loop it then which then becomes equivalent to using `io.ReadAll`/`io.Copy` based approach.

PR #203, opened 2025-08-08
refactor(mysecret): parameterize store so and mock for tests

### LINKED ISSUE #200: Basic operations: set/get/rm/list

joe0BAB:
Refactors so the store is parameterized and can be replaced with a mock for tests. Adds a basic unit test for `mysecret set` command.

(I'll fix and extend this to cover all cases in follow up PR, wanted to keep this PR simple and just get started setting up tests at all)


mysecret/command_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/docker/secrets-engine/internal/secrets"
+	"github.com/docker/secrets-engine/mysecret/service"
+	"github.com/docker/secrets-engine/store"
+)
+
+type mockStoreOption func(m *mockStore)
+
+type mockStore struct {
```
Benehiko, line 20: non-blocking; i wonder if you can't just re-use the mocks package in the `store` module.

Benehiko (approved):

mysecret/command_test.go
joe0BAB: I had this initially but then also had to mock failure which feels out of scope for your general mock. For the tests I'll have to keep extending the mock to also mimic any kind of failure behavior.

PR #205, opened 2025-08-11
refactor: export ID field in ErrInvalidID

joe0BAB:
Tiny refactoring so we can use `ErrorIs` in tests on `ErrInvalidID`.


joe0BAB:
For CI failures, see #206 that fixes it.

Benehiko (approved):

PR #217, opened 2025-08-11
feat(mysecret): boilerplate code to add mysecret as internal engine  plugin

### LINKED ISSUE #212: MySecret Plugin
Plugin that allows to fetch secrets set via `docker mysecret` through the secrets engine.

joe0BAB:
Plumping and boilerplate to setup mysecret as internal plugin part of the engine daemon.


engine/daemon/internal/mysecret/mysecret.go
```diff
@@ -0,0 +1,24 @@
+package mysecret
+
+import (
+	"context"
+
+	"github.com/docker/secrets-engine/engine"
+	"github.com/docker/secrets-engine/internal/secrets"
+)
+
+type mysecretPlugin struct{}
```
Benehiko, line 10: could we assert the interface it is relying on?

Benehiko (approved):

engine/daemon/internal/mysecret/mysecret.go
joe0BAB: yes, let me do in the next / follow up PR

PR #219, opened 2025-08-11
feat(daemon): implement mysecret plugin

### LINKED ISSUE #212: MySecret Plugin

joe0BAB:
Implements mysecret plugin and adds it as builtin plugin to the engine.


engine/go.mod
```diff
@@ -9,24 +9,35 @@ replace github.com/docker/secrets-engine => ../
...
 replace github.com/docker/secrets-engine/plugin => ../plugin
 
+replace github.com/docker/secrets-engine/store => ../store
+
 require (
 	connectrpc.com/connect v1.18.1
 	github.com/cenkalti/backoff/v5 v5.0.3
 	github.com/docker/secrets-engine v0.0.7
 	github.com/docker/secrets-engine/client v0.0.7
+	github.com/docker/secrets-engine/mysecret v0.0.0-00010101000000-000000000000
```
joe0BAB, line 23: Should we do another release before the PR so we can get rid of this? (mysecret hasn't been released yet)

Benehiko: nothing is importing engine yet, so should be fine

Makefile
```diff
@@ -82,7 +82,7 @@ keychain-unit-tests:
 	CGO_ENABLED=1 go test -v $$(go list ./store/keychain/...)
 
 engine-unit-tests:
-	CGO_ENABLED=0 go test -v $$(go list ./engine/...)
+	CGO_ENABLED=1 go test -v $$(go list ./engine/...)
```
Benehiko: maybe we should be explicit about CGO, let's remove the `CGO_ENABLED=1` here and move it to the workflow where it is used. We can even do a separate PR for it (let's cherry pick this commit and merge it https://github.com/docker/secrets-engine/pull/221/commits/9f4cae52dd07157eba00ec54165f84a29da3cde1)

Benehiko: https://github.com/docker/secrets-engine/pull/222

Benehiko (approved):

PR #221, opened 2025-08-12
engine: add docker-auth built-in plugin

Benehiko:
Add the github.com/docker/docker-auth as a built-in plugin of the engine.


engine/daemon/internal/dockerauth/auth.go
```diff
@@ -0,0 +1,78 @@
...
+		expiresAt, err = time.Parse(time.RFC3339, v)
+		if err != nil {
+			return secrets.EnvelopeErr(request, err), err
+		}
+	}
+
+	return secrets.Envelope{
+		ID:         request.ID,
+		Value:      val,
+		Provider:   serviceName,
```
joe0BAB, line 57: FYI, the engine will overwrite this with the provider string the plugin registers for

joe0BAB (approved):
LGTM! (I think we somehow should reference in the issue created for the secret type and it's marshaling that once that's done latest we should add tests to this plugin here as well)

engine/daemon/internal/dockerauth/auth.go
Benehiko: ah i see, we should somehow make this more clear then for third-party integrators.

PR #229, opened 2025-08-12
fix(ci): for local linting read GITHUB_TOKEN from env

joe0BAB:


Benehiko (changes requested):
We should rely on ssh for local linting instead.

Benehiko:
done in #239

joe0BAB:
Re-opening, as merging that PR doesn't interfere with SSH based local git setup / is independent of that and basically just says "you don't need any local git setup to do the linting as long as you provide the GITHUB_TOKEN".

joe0BAB:
Never mind, I missed that you already have done this.

PR #239, opened 2025-08-21
fix: `make lint` on local should use ssh keys

Benehiko:
On the CI we have access to a GitHub token which can pull in private repositories. On local environments we should stick with convention and use SSH keys instead. It's more secure and doesn't require setting bespoke environment variables, it simply relies on what is publicly documented and recommended by GitHub (see https://docs.github.com/en/authentication/connecting-to-github-with-ssh)


Dockerfile
```diff
@@ -11,7 +11,7 @@ FROM --platform=${BUILDPLATFORM} tonistiigi/xx AS xx
 FROM --platform=${BUILDPLATFORM} crazymax/osxcross:${OSXCROSS_VERSION}-alpine AS osxcross
 
 FROM --platform=${BUILDPLATFORM} golang:${GO_VERSION}-alpine AS gobase
-RUN apk add --no-cache findutils build-base git
+RUN apk add --no-cache findutils build-base git
```
joe0BAB: nit: drop this from the PR

Makefile
```diff
@@ -51,7 +51,13 @@ format: ## Format code
...
 
 lint: multiarch-builder ## Lint code
-	@docker buildx build $(DOCKER_BUILD_ARGS) --secret id=GH_TOKEN --pull --builder=$(BUILDER) --target=lint --platform=$(LINT_PLATFORMS) .
+	@set -e; \
+	flags="--ssh default"; \
+	if [ -n "$$GH_TOKEN" ]; then \
+		flags="--secret id=GH_TOKEN,env=GH_TOKEN"; \
+	fi; \
+	echo "Using build flags: $$flags"; \
+	docker buildx build $(DOCKER_BUILD_ARGS) $$flags --pull --builder=$(BUILDER) --target=lint --platform=$(LINT_PLATFORMS) .
```
joe0BAB: non blocking: is there any benefit of `$$` vs `$()`? Should we try to be consistent?

joe0BAB (approved):
LGTM! (some small optional nits)

PR #240, opened 2025-08-22
feat(sdk): resolver v2 interface (pattern based queries)

### LINKED ISSUE #220: Resolver Interface v2: allow pattern queries
The current resolver interface, i.e., the interface to query any secrets from the engine, only allows simple get requests of the form `Get(ID, provider)` but doesn't take advantage of the realm system baked into IDs and patterns. Additionally, plugins/providers have to define a pattern which acts as a filter on which queries get forwarded to that plugin.

# Proposal
1. As any ID is a pattern, we could expand this to directly allow querying through patterns.
2. To avoid confusing between what patterns are used for, we should rename the pattern defined on a plugin as filter as the main goal of defining patterns on plugins is performance, pre-filter which requests get forwarded to which plugin but the actual pattern match still has to be done within each plugin.
3. Allow returning multiple secrets: Right now the engine will return the first match. However, the engine in general has not enough context to decide which potential match is the best/right one. That should be sorted out on the client where more context is available.

# Reasoning/Motivation
1. Allow building cross plugin business logic. Without pattern queries, the match/realm logic needs to be done per plugin. With pattern queries, plugins can be treated as "dumb sources" that only provide values and business logic can be created that combines multiple plugins.

## Example 1: `*` match providers in realm paths

Let's say the way to retrieve an auth user token is:

```
auth/alice/my-private-repo
```

Now let's say there are two environments, staging and prod with two respective registries one hosted on docker hub and the other on JFrog. For staging, her token is stored in local keychain, but for prod it's in Vault.

With `Get(ID, provider)` style queries, everything needs to be specified up-front, ie, the client has to know about Vault vs local keychain and eg query `auth/alice/my-private-repo?provider=vault` or `vault/alice/my-private-repo` if we decided to define a component schema that encodes the provider into the top level realm.

With `Get(pattern, provider)` style queries, this could be generalized into `*/alice/my-private-repo`.

## Example 2: `*` matching registries

If we were to add the registry into above example, eg, let's say we define the schema:
```
[<provider>/]<registry>/<user>/<repo> 
```
things become a bit more limited. As there's no `registry` field in the `request` struct / the registry has no direct mapping to plugins/providers, with `Get(ID, provider)` style queries the `registry` component always has to be known ahead / explicitly set as part of the ID (alternatively one could argue, plugins if no registry is provided in the ID query they fallback to something, but then every plugin/provider has to re-implement that same implicit convention which is rather fragile).

For pattern based queries `Get(pattern, provider)`, this can be changed to eg `vault/*/alice/my-private-repo`, `*/*/alice/my-private-repo`.

# Implementation

## Sub-pattern check
In order for the plugin filter logic to work as before, we need a `Pattern.IsSubPattern(pattern)` function that checks if the matches of pattern A are a subset of the matches of pattern B.

## Interface Change
Before:

```go
type Resolver interface {
	GetSecret(ctx context.Context, request Request) (Envelope, error)
}

type Request struct {
	ID ID `json:"-"`

	// Provider can be optionally specified to restrict the resolver
	// to a particular provider stack.
	Provider    string    `json:"-"`
	ClientID    string    `json:"-"`
	RequestedAt time.Time `json:"-"`
}
```

After:

```go
type Resolver interface {
	GetSecret(ctx context.Context, request Request) ([]Envelope, error)
}

type Request struct {
	Pattern Pattern `json:"-"`

	// Provider can be optionally specified to restrict the resolver
	// to a particular provider stack.
	Provider    string    `json:"-"`
	ClientID    string    `json:"-"`
	RequestedAt time.Time `json:"-"`
}
```

joe0BAB:
This is the main chunk of upgrading the resolver interface to allow patterns. It replaces `ID` with `Pattern` in the request and the return value to a list of envelopes. It also removes the error field on the envelope type. There are still some TODOs left for a follow up PR (add logging at certain places and further change the interface to `GetSecrets(context, pattern)`).


engine/daemon/internal/dockerauth/auth.go
```diff
@@ -20,19 +21,35 @@ type dockerAuthPlugin struct {
...
-		return secrets.EnvelopeErr(request, err), err
+		return secrets.EnvelopeErrs(err), err
+	}
+
+	var errList []error
+	var result []secrets.Envelope
+	for id, value := range list {
+		s, err := unpackValue(id, value)
+		errList = append(errList, err)
+		result = append(result, s)
```
joe0BAB: The alternative would be to fail fast and abort on the first error.

x/secrets/secrets.go
```diff
@@ -14,7 +14,7 @@ var (
 )
 
 type Request struct {
-	ID ID `json:"-"`
+	Pattern Pattern `json:"-"`
```
joe0BAB, line 17: Main change (and the rest in this file)

engine/daemon/internal/dockerauth/auth.go
```diff
@@ -20,19 +21,39 @@ type dockerAuthPlugin struct {
...
-		return secrets.EnvelopeErr(request, err), err
+		return nil, err
+	}
+
+	var errList []error
+	var result []secrets.Envelope
+	for id, value := range list {
+		s, err := unpackValue(id, value)
+		if err != nil {
+			errList = append(errList, err)
```
Benehiko: i don't think we should even add this to an error list.
let's just log and continue.

engine/daemon/internal/dockerauth/auth.go
```diff
@@ -20,19 +21,39 @@ type dockerAuthPlugin struct {
...
+			// TODO: log error
+			continue
+		}
+		result = append(result, *s)
+	}
+
+	if len(result) == 0 && len(errList) > 0 {
```
Benehiko: if len(result) == 0 we just return NotFound error.

engine/daemon/internal/mysecret/mysecret.go
```diff
@@ -16,22 +16,38 @@ type mysecretPlugin struct {
...
+		return nil, err
 	}
-	impl, ok := s.(*service.MyValue)
+
+	var errList []error
+	var result []secrets.Envelope
+	for id, value := range list {
+		s, err := unpackValue(id, value)
+		if err != nil {
+			errList = append(errList, err)
```
Benehiko: same here, we shouldn't need to add all the errors to a list.

engine/resolver.go
```diff
@@ -15,37 +14,41 @@ type regResolver struct {
...
-	var err error
-	if len(errs) == 0 {
-		err = fmt.Errorf("secret %q: %w", req.ID, secrets.ErrNotFound)
-	} else {
-		err = errors.Join(errs...)
+	if len(results) > 0 {
+		return results, nil
+	}
+
+	if len(results) == 0 && len(errs) == 0 {
```
Benehiko: ignore the errors here, just check len of results

joe0BAB:
Addressed the behavior change of not forwarding internal errors but only a not found error in 64e945bc3ea5e270878aabe666b81155d714ed5a

Benehiko (approved):

PR #241, opened 2025-08-22
refactor: re-use the resolver service handler type mapping

### LINKED ISSUE #220: Resolver Interface v2: allow pattern queries

joe0BAB:
This is prep work to reduce the upcoming refactoring needed to complete #220 .

With #240 , this handler code I'm moving around here will become a bit more complex. So wanted to not have duplication plus assure it's tested.


engine/resolver.go
```diff
@@ -6,57 +6,16 @@ import (
...
-		Id:         proto.String(envelope.ID.String()),
-		Value:      envelope.Value,
-		Provider:   proto.String(envelope.Provider),
-		Version:    proto.String(envelope.Version),
-		Error:      proto.String(envelope.Error),
-		CreatedAt:  timestamppb.New(envelope.CreatedAt),
-		ResolvedAt: timestamppb.New(envelope.ResolvedAt),
-		ExpiresAt:  timestamppb.New(envelope.ExpiresAt),
-	}.Build()), nil
-}
```
joe0BAB, line 51: The main change is deduplicating this and adding tests.

joe0BAB:
First test run was flaky. Tried to dig into and noticed a potential race condition in the test setup so added f013ec360ec890485c8002193668b9cced37bf9c

Benehiko (approved):

PR #267, opened 2025-09-01
feat: release automation

### LINKED ISSUE #265: Automate sub module release process

joe0BAB:
Implements #265 : Adds a new command to automate creating release PRs.

To use the new command, run `make bump MOD=x` to create a new patch release of module `x`. It will automatically propagate through the version bump of `x` to all internal modules and create new versions for those which depend on it (you still need to create a branch/PR and merge all changes). Automatically creates commits for the go.mod changes and git tags. Use `go run ./x/release` for more fine granular control (eg to do minor/major releases).


x/release/command.go
```diff
@@ -0,0 +1,251 @@
...
+			}
+			return helper.BumpIterative(mod, level, data, &projectFS{opts: opts, logger: logging.NewDefaultLogger("")})
+		},
+	}
+
+	flags := bump.Flags()
+	flags.BoolVar(&opts.dryRun, "dry", false, "Dry run: Only log all steps, no actual changes will be made.")
+	flags.BoolVar(&opts.skipGit, "skip-git", false, "Skip git operations: Useful to preview only the go.mod changes.")
+	flags.BoolVar(&opts.minor, "minor", false, "Do a minor release (downstream dependencies will only get a patch release bump).")
+	flags.BoolVar(&opts.major, "major", false, "Do a major release (downstream dependencies will only get a patch release bump).")
```
Benehiko: since we cannot use both, perhaps we could make this a string instead of a bool e.g. `--level minor` then you can have an autocomplete of the supported options: `minor` and `major`.

Benehiko: i'd rather be explicit here and also specify `patch` as an option instead of just automatically falling back to it.

x/release/command.go
```diff
@@ -0,0 +1,251 @@
...
+	if opts.minor && opts.major {
+		return -1, fmt.Errorf("cannot use both --minor and --major at the same time")
+	}
+	if opts.major {
+		return helper.Major, nil
+	}
+	if opts.minor {
+		return helper.Minor, nil
+	}
+	return helper.Patch, nil
```
Benehiko: if opt isn't specified rather error instead of falling back to patch by default.

x/release/main.go
```diff
@@ -0,0 +1,23 @@
+package main
+
+import (
+	"context"
+	"os/exec"
+)
+
+func main() {
+	ctx := context.Background()
```
Benehiko: this context doesn't do much, we should handle a SIGINT/SIGTERM here

x/release/helper/release.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	if !needsCommit {
+		return nil
+	}
+	return i.GitCommit("chore: bump " + tag)
+}
+
+type RepoData map[string]ModData
+
+func (d *RepoData) AddMod(mod string, data FutureVersions, deps []string) {
```
Benehiko: let's not use a pointer for `RepoData` receiver. It forces us to de-reference, which looks ugly.

x/release/helper/release.go
```diff
@@ -0,0 +1,213 @@
...
+	(*d)[dep] = before
+}
+
+func (d *RepoData) getOrInit(mod string) ModData {
+	result, ok := (*d)[mod]
+	if !ok {
+		return ModData{downstreams: map[string]struct{}{}}
+	}
+	return result
+}
```
Benehiko: we should maybe just init the map before using it, then we can get rid of this function and get rid of the pointer receiver.

x/release/helper/release.go
```diff
@@ -0,0 +1,213 @@
...
+
+func (d *RepoData) GetDependencyCount(mod string) int {
+	data, ok := (*d)[mod]
+	if !ok {
+		return 0
+	}
+	return len(data.dependencies)
+}
+
+func (d *RepoData) RmMod(mod string) {
```
Benehiko:
```suggestion
func (d *RepoData) RemoveModule(mod string) {
```

x/release/helper/release.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+}
+
+type Opt func(f *FutureVersions)
+
+func NewFutureVersions(latest string, opts ...Opt) (*FutureVersions, error) {
+	v := semver.Canonical(latest)
+	if v == "" {
+		return nil, fmt.Errorf("not a canonical version: %s", latest)
+	}
```
Benehiko: what about cases where we haven't published a tag yet, then `latest` might be `(devel)` which means it does not conform to semver. 

See here: https://github.com/docker/secrets-engine/blob/main/engine/daemon/main.go#L22-L26

x/release/helper/release.go
```diff
@@ -0,0 +1,213 @@
...
+	return func(v *FutureVersions) {
+		v.patch += value
+		v.minor += value
+		v.major += value
+	}
+}
+
+type Opt func(f *FutureVersions)
+
+func NewFutureVersions(latest string, opts ...Opt) (*FutureVersions, error) {
```
Benehiko: I still don't understand why we cannot just pass in the future version we want by level. e.g. I have version `v0.1.2` - get me the future `Patch` version -> `v0.1.3` is the result.

Why do we need to construct a struct, increment all values and then select them later? Why not just make this a utility function with one purpose of returning the correct future version based on a given level?

x/release/command.go
```diff
@@ -0,0 +1,251 @@
...
+
+	"github.com/spf13/cobra"
+	"golang.org/x/mod/modfile"
+
+	"github.com/docker/secrets-engine/x/logging"
+	"github.com/docker/secrets-engine/x/release/helper"
+)
+
+type Config struct {
+	versionExtraAllowList []string
```
Benehiko: The `Extra` is called a prerelease or build release https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions. Maybe we should name it accordingly to avoid confusion? Also how will others configure this if the package is imported by a third party?

Let's rename this to `EnableModulesWithPreReleaseVersion` (feel free to suggest an alternative)?

x/release/command.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	m.logger.Printf(prefix+"git commit: '%s'", commit)
+	if m.dryRun || m.skipGit {
+		return nil
+	}
+	return gitCommit(commit)
+}
+
+func newRepoData(versionExtraAllowList []string) (helper.RepoData, error) {
+	entries, err := os.ReadDir(".")
```
Benehiko: This might be a bit excessive, but let's prevent following symlinks by wrapping this inside `os.Root` - that way, only the current directory is available to us when opening directories / files.

x/release/command.go
```diff
@@ -0,0 +1,251 @@
...
+	if err != nil {
+		return nil, err
+	}
+	modules := helper.RepoData{}
+	for _, entry := range entries {
+		if !entry.IsDir() {
+			continue
+		}
```
Benehiko, line 165: Is it intentional that we are excluding mod files in the root directory? AFAIK this will skip all files and only open directories (which may contain files such as "go.mod").

x/release/command.go
```diff
@@ -0,0 +1,251 @@
...
+	for _, require := range f.Require {
+		if strings.HasPrefix(require.Mod.Path, project) {
+			deps = append(deps, strings.TrimPrefix(require.Mod.Path, project+"/"))
+		}
+	}
+	return deps, nil
+}
+
+func getLatestVersion(modName string) (string, error) {
+	cmd := exec.Command("git", "tag", "-l", modName+"/v*", "--sort=-v:refname")
```
Benehiko, line 204: we could use go-git here instead of the git CLI.
https://pkg.go.dev/github.com/go-git/go-git/v5#Repository.CreateTag

We used it in copilot agent to clone repo's https://github.com/docker/copilot-agent/blob/main/pkg/git/git.go

x/release/command.go
joe0BAB: let me refactor to using a custom type then, that's the cleanest solution

x/release/helper/release.go
joe0BAB: don't fully understand this, it's a map of maps, how would you initialize this properly?

x/release/command.go
joe0BAB: this code is no longer needed when refactoring to a custom type

x/release/helper/release.go
joe0BAB: Initially this was simpler for debugging purposes. But by now it smells a bit. Let me refactor..

x/release/command.go
joe0BAB: I was reluctant to pull in such a heavy dependency for this (particular as in our case this is only CI helper stuff and not main concern of this repo)

x/release/command.go
joe0BAB: Could we do this in a follow up PR? We don't need it yet.

Benehiko: i think you can just remove the if statement and it _should_ work.

Makefile
```diff
@@ -139,6 +139,28 @@ mod:
...
+export HELP_BUMP
+
+bump: MOD?=
+bump:
+	@if [ -z "$(MOD)" ]; then \
+		echo "$$HELP_BUMP"; \
+		exit 1; \
+	fi
+	git fetch --tags
+	go run ./x/release $(MOD)
```
Benehiko, line 162: We don't support the release flag here?

Benehiko (approved):
Much better!

Makefile
joe0BAB: Make targets get very fiddly with flags. But that's why I added the help text -> for more non trivial use cases, just use the release helper binary directly

PR #271, opened 2025-09-08
feat: go mod internal version bump helper

### LINKED ISSUE #265: Automate sub module release process

joe0BAB:
More helper functions extracted from #267 .


x/release/release.go
```diff
@@ -3,18 +3,92 @@ package main
...
 
+type fs interface {
+	gitTag(tag string) error
+	gitCommit(msg string) error
+	bumpModfile(path, module, version string) error
+	makeMod() error
+}
+
+func bumpModule(mod string, data modData, i fs) error {
+	fmt.Printf("bumping module %s\n", mod)
```
Benehiko: we shouldn't use fmt here, rather use the logger

x/release/release.go
```diff
@@ -3,18 +3,92 @@ package main
...
+	return result
+}
+
+type modData struct {
+	downstreamDeps map[string]struct{}
+	dependencies   map[string]struct{}
+	futureVersions
+}
+
+const projectPathPrefix = "github.com/docker/secrets-engine/"
```
Benehiko: i guess we might want to re-use this package internally? let's not hard-code the project values here

x/release/release.go
```diff
@@ -3,18 +3,92 @@ package main
...
+	}
+	before := d.get(mod)
+	before.dependencies = dependencies
+	(*d)[mod] = before
+}
+
+func (d *repoData) addDownstreamDep(mod, dep string) {
+	before := d.get(dep)
+	before.downstreamDeps[mod] = struct{}{}
+	(*d)[dep] = before
```
Benehiko: this looks a bit wild

x/release/release.go
```diff
@@ -3,18 +3,92 @@ package main
...
+	result, ok := (*d)[mod]
+	if !ok {
+		return modData{downstreamDeps: map[string]struct{}{}}
+	}
+	return result
+}
+
+type modData struct {
+	downstreamDeps map[string]struct{}
+	dependencies   map[string]struct{}
```
Benehiko: unsure what "downstreamDeps" are and what the difference is between that and "dependencies".

x/release/release.go
```diff
@@ -3,18 +3,92 @@ package main
...
+	(*d)[mod] = before
+}
+
+func (d *repoData) addDownstreamDep(mod, dep string) {
+	before := d.get(dep)
+	before.downstreamDeps[mod] = struct{}{}
+	(*d)[dep] = before
+}
+
+func (d *repoData) get(mod string) modData {
```
Benehiko: feels weird that `get` would never explicitly fail.

x/release/release.go
```diff
@@ -3,18 +3,92 @@ package main
...
 	minor string
 	major string
 }
 
+func versionAppend(v futureVersions, value string) futureVersions {
+	v.patch += value
+	v.minor += value
+	v.major += value
+	return v
+}
```
Benehiko: unsure of how this is used, but feels obscure.

x/release/release.go
```diff
@@ -3,18 +3,92 @@ package main
...
+	gitTag(tag string) error
+	gitCommit(msg string) error
+	bumpModfile(path, module, version string) error
+	makeMod() error
+}
+
+func bumpModule(mod string, data modData, i fs) error {
+	fmt.Printf("bumping module %s\n", mod)
+
+	tag := mod + "/" + data.patch
```
Benehiko: so the tag will only contain the patch version?

x/release/release.go
joe0BAB: let's say we have a module x with dependencies y and z. Then x would be a downstream dependency for module y and z (and get added to their `downstreamDeps` maps). It's the reverse mapping of dependencies.

x/release/release.go
joe0BAB: rewrote this a bit so any version can be selected

x/release/release.go
joe0BAB: It's an internal helper. Renamed to `getOrInit`.

x/release/release.go
joe0BAB: Refactored this to become an optional parameter of the new function.

x/release/helper/release.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+}
+
+type Opt func(f *FutureVersions)
+
+func NewFutureVersions(latest string, opts ...Opt) (*FutureVersions, error) {
+	v := semver.Canonical(latest)
+	if v == "" {
+		return nil, fmt.Errorf("not a canonical version: %s", latest)
+	}
```
Benehiko, line 120: what if current latest is a dev version (it's usually what the Go runtime makes up if there is no tag associated). 
See https://github.com/docker/secrets-engine/blob/main/engine/daemon/main.go#L23-L26

x/release/helper/release.go
```diff
@@ -0,0 +1,150 @@
...
+	if err := i.BeforeCommit(); err != nil {
+		return err
+	}
+	return i.GitCommit("chore: bump " + tag)
+}
+
+type RepoData map[string]ModData
+
+func (d *RepoData) AddMod(mod string, data FutureVersions, deps []string) {
```
Benehiko, line 48: We don't need to make a pointer receiver of the RepoData type. Leave it as a non-pointer then your dereferencing wouldn't be necessary.

x/release/helper/release.go
```diff
@@ -0,0 +1,150 @@
...
+	downstreams  map[string]struct{}
+	dependencies map[string]struct{}
+	FutureVersions
+}
+
+type FutureVersions struct {
+	patch string
+	minor string
+	major string
+}
```
Benehiko, line 94: i still get confused by this - i understand now that `patch` is the full version but with only the patch incremented. But it feels weird that we are pre-calculating it and then selecting from it instead of just calculating a singular value upon request. 

I keep second guessing myself when reading code like on line 31 https://github.com/docker/secrets-engine/pull/271/files#diff-e05bae9af44e9497d3839a75a4480be70aef9242eaabc26cb99fffea730bd09fR31

"is this only printing the patch part of the version? e.g `v1.1.0` -> `0`"

Benehiko (commented):
It looks much better, but i think the `FutureVersion` type is confusing. Feels like we are doing so much here when we just need to calculate a singular string based on input.

PR #272, opened 2025-09-09
feat: iterative release resolver helper

### LINKED ISSUE #265: Automate sub module release process

joe0BAB:
Helper function to iteratively bump and resolve our internal modules/dependencies.


x/release/release.go
```diff
@@ -68,6 +111,24 @@ func (d *repoData) get(mod string) modData {
 	return result
 }
 
+func (d *repoData) nrDeps(mod string) int {
```
Benehiko: what does `nrDeps` mean?

x/release/release.go
```diff
@@ -68,6 +111,24 @@ func (d *repoData) get(mod string) modData {
...
+	data, ok := (*d)[mod]
+	if !ok {
+		return 0
+	}
+	return len(data.dependencies)
+}
+
+func (d *repoData) rmDep(mod string) {
```
Benehiko: Let's make the name explicit
```suggestion
func (d *repoData) removeDependency(mod string) {
```

x/release/release.go
```diff
@@ -68,6 +111,24 @@ func (d *repoData) get(mod string) modData {
...
+	delete(*d, mod)
+	after := repoData{}
+	for k, v := range *d {
+		delete(v.dependencies, mod)
+		after[k] = v
+	}
+	*d = after
```
Benehiko: this code feels a bit clunky. I'm wondering if we can get away with just the delete if `v.dependencies` are referenced directly from the map through `k` instead of using `v` (the copy)?

joe0BAB: tests still are green when removing it, then let's simplify!

PR #274, opened 2025-09-10
feat: merge New/Run functions to make running the engine stateless

### LINKED ISSUE #273: Prepare Engine for Integration into Docker Desktop
We need to adjust the interface a bit to make integration with DD easier and comply with conventions there.

joe0BAB:
By now there's not much point anymore in having a dedicated `New` method to create an engine. We should merge this with the `Run` function.


engine/adaptation.go
```diff
@@ -153,8 +141,15 @@ func WithMaxTries(maxTries uint) Option {
...
 
-// New creates a new NRI Runtime.
-func New(name, version string, opts ...Option) (Engine, error) {
+// WithOnUp set a callback that gets called once the engine is ready to accept requests.
+func WithOnUp(cb func()) Option {
+	return func(r *config) error {
+		r.upCb = cb
+		return nil
+	}
+}
```
Benehiko:
```suggestion
// WithAfterHealthyHook set a callback that gets called once the engine is ready to accept requests.
func WithAfterHealthyHook(hook func(ctx context) error) Option {
	return func(r *config) error {
		r.upCb = cb
		return nil
	}
}
```

we should perhaps treat the hook as synchronous? Meaning the engine should propagate any errors produced by the hook?

Benehiko: in my experience having hooks called async makes things really difficult to debug and i don't think the use case calls for that yet.

engine/adaptation.go
```diff
@@ -77,12 +70,7 @@ type config struct {
 	enginePluginsDisabled  bool
 	logger                 logging.Logger
 	maxTries               uint
-}
-
-type adaptation struct {
-	config
-
-	m sync.Mutex
+	upCb                   func() error
```
Benehiko:
```suggestion
	upCb                   func(context.Context) error
```

It would be best to pass along the engine context - just in case there's a long running operation inside the callback.

Benehiko (approved):

PR #279, opened 2025-09-11
fix(engine): fix data race on shutdown

### LINKED ISSUE #273: Prepare Engine for Integration into Docker Desktop

### LINKED ISSUE #280: Run unit tests with `-race` flag and fix all data race issues

joe0BAB:
Fixes a data race in the engine and enables `-race` flag for unit tests (except for engine, need to fix more data races first).


engine/engine.go
```diff
@@ -51,26 +51,23 @@ func newEngine(ctx context.Context, cfg config) (engine, error) {
...
 	}()
 	return &engineImpl{
 		reg: reg,
 		close: sync.OnceValue(func() error {
 			defer cfg.listener.Close()
-			select {
-			case <-done:
-				return serverErr
-			default:
-			}
```
joe0BAB, line 68: I misunderstood how http.Server.Shutdown works. We can always call it regardless if some http.Server.Serve crashed. The only limitation is that we shouldn't call it more than once (which already prevent anyway)

Benehiko (approved):

PR #282, opened 2025-09-12
fix(engine): make cmd watcher synchronous to avoid data races later

joe0BAB:


Benehiko (commented):
Wouldn't this increase data races because we are splitting the Run function (which was blocking?)

joe0BAB:
The current problem this aims to solve is that we run a cleanup close call on the file descriptors inherited to the child process in the parent process before the child has started (that's the race issue, that there's no guarantee it has started yet). We can solve this two ways:
- making the cmd launch wrapper synchronous (what this PR is doing)
- add a callback "onStarted" parameter and trigger the parent cleanup from there

As we anyway need to wait for the child process to have been started later, there's no advantage in keeping the launch wrapper async. 

The cmd.Run is only called inside this wrapper. The interface/generic stuff is only for being able to write unit tests and emulating specific scenarios. So no, I don't think so (or maybe you meant something else that I missed)

Benehiko (approved):

PR #286, opened 2025-09-15
fix(engine): iterating over all plugins is concurrency safe

### LINKED ISSUE #280: Run unit tests with `-race` flag and fix all data race issues

joe0BAB:
The previous implementation of `GetAll() []runtime` was not actually concurrency safe as while accessing the slice inside `GetAll()` was protected by mutex, the returned slice would be used in unprotected ways. One option to fix this is to implement an `Iterator` interface with a `Iterator.Next()` method that mutex protected gets the next `runtime`.


engine/manager.go
```diff
@@ -13,13 +13,31 @@ type removeFunc func()
 
 type registry interface {
 	Register(plugin runtime) (removeFunc, error)
-	GetAll() []runtime
+	Iterator() iterator
+}
+
+type iterator interface {
+	Next() runtime
```
Benehiko: i think there's a built-in type for this https://pkg.go.dev/iter#Seq2

engine/manager.go
```diff
@@ -51,18 +69,50 @@ func (m *manager) sort() {
...
+	m.m.Lock()
+	defer m.m.Unlock()
+	startIdx := -1
+	if len(m.plugins) > 0 {
+		startIdx = 0
+	}
+	result := &visitor{index: startIdx, m: m}
+	m.visitors[result] = struct{}{}
+	return result
+}
```
Benehiko: not sure why we are implementing this from scratch - there is a built-in for this https://pkg.go.dev/iter#Seq2

Benehiko (approved):
LGTM

PR #289, opened 2025-09-15
chore: bump Go version to 1.25 (revert 26874cf)

Benehiko:
Bump Go to 1.25 again by reverting 26874cf.


joe0BAB:
Could we upgrade to 1.25.1 directly?

Benehiko:
There doesn't seem to be anything of significance for us there: https://github.com/golang/go/issues?q=milestone%3AGo1.25.1+label%3ACherryPickApproved

cmd/nri-plugin/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/docker/secrets-engine/nri-plugin
 
-go 1.24.6
+go 1.24.3
```
joe0BAB, line 3: Why do we downgrade here? 🤔

joe0BAB: ah never mind (we are rolling back 🤦 )

joe0BAB (approved):

cmd/nri-plugin/go.mod
Benehiko: yes, although, we should fix this :laughing:

PR #292, opened 2025-09-17
feat: cross-platform flock on filestore

Benehiko:
This patch adds support for cross-platform file locks including a recovery pattern for stale locks. 

It behaves as follows:

- It uses standard OS APIs to create an advisory lock on a temporary file. 
- The file gets truncated to update its last modification time.
- If the file is still locked and the modification time was more than 30 seconds ago, it will be removed and re-created (at least on Unix systems it can. On Windows it will fail).
- tryLock has retry logic, and will retry for a max of 100ms starting at 10ms delays and exponentially waiting longer between retries.
- attemptLock only retries once more after a successful recovery
- any errors on recovery will abort attemptLock


store/posixage/flock.go
```diff
@@ -0,0 +1,127 @@
...
+
+var (
+	ErrLockUnsuccessful   = errors.New("store is locked")
+	ErrUnlockUnsuccessful = errors.New("could not unlock store")
+)
+
+const lockFileName = ".posixage.lock"
+
+// unlockFunc is the callback function returned by [lockFile]
+// it should always be called inside a defer.
```
joe0BAB: alternatively we could remove the comment and use an internal package to make sure implementation details that shouldn't be used separately can't be used separately by design

Benehiko: they are already internal to the posixage package

store/posixage/flock.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, err
+	}
+	// truncate to update the modtime to signal to other processes that the
+	// current lock is valid so they don't attempt a recovery on it.
+	_ = fl.Truncate(0)
+
+	return sync.OnceValue(func() error {
+		return unlockFile(fl)
+	}), nil
+}
```
joe0BAB:
```suggestion
func attemptLock(ctx context.Context, root *os.Root, exclusive bool) (unlockFunc, error) {
	var fl *os.File
	for {
		var err error
		fl, err = openFile(root)
		if err != nil {
			return nil, err
		}
		if err := tryLockFile(ctx, fl, exclusive); err != nil {
			break
		}
		if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
			// maybe there's a better way to guarantee we don't end in an infinite loop?
			return nil, err
		}

		// lock was unsuccessful so let's retry
		if recoverErr := recoverLock(root, fl); recoverErr != nil {
			// return on recovery failed.
			// perhaps the file is still locked and not older than 30 seconds?
			// maybe a permission error prevented it from being removed?
			return nil, errors.Join(err, filterRecoverError(recoverErr))
		}
	}

	// truncate to update the modtime to signal to other processes that the
	// current lock is valid so they don't attempt a recovery on it.
	_ = fl.Truncate(0)

	return sync.OnceValue(func() error {
		return unlockFile(fl)
	}), nil
}
```
why not just use a for loop?

store/posixage/flock.go
```diff
@@ -0,0 +1,127 @@
...
+// 30 seconds. This assumes no valid process would hold the lock that long.
+func attemptLock(ctx context.Context, root *os.Root, exclusive bool) (unlockFunc, error) {
+	fl, err := openFile(root)
+	if err != nil {
+		return nil, err
+	}
+
+	lockCtx, lockCtxCancel := context.WithTimeout(ctx, time.Millisecond*100)
+	defer lockCtxCancel()
+	err = tryLockFile(lockCtx, fl, exclusive)
```
joe0BAB:
```suggestion
	err = tryLockFile(lockCtx, fl, exclusive)
```
why not push that down into `tryLockFile`?

store/posixage/flock_unix.go
```diff
@@ -0,0 +1,56 @@
...
+			return false, err
+		}
+		return true, nil
+	}, backoff.WithBackOff(ep))
+	if err != nil {
+		return errors.Join(ErrLockUnsuccessful, err)
+	}
+
+	return nil
+}
```
joe0BAB:
```suggestion
func tryLockFile(ctx context.Context, f *os.File, exclusive bool) error {
	lockCtx, lockCtxCancel := context.WithTimeout(ctx, time.Millisecond*100)
	defer lockCtxCancel()
	
	ep := backoff.NewExponentialBackOff()
	ep.InitialInterval = time.Millisecond * 10
	ep.MaxInterval = time.Millisecond * 100
	_, err := backoff.Retry(lockCtx, func() (bool, error) {
		if err := lockFile(f, exclusive); err != nil {
			return false, err
		}
		return true, nil
	}, backoff.WithBackOff(ep))
	if err != nil {
		return errors.Join(ErrLockUnsuccessful, err)
	}

	return nil
}

func lockFile(f *os.File, exclusive bool) error {
	flag := unix.LOCK_NB
	if exclusive {
		flag |= unix.LOCK_EX
	} else {
		flag |= unix.LOCK_SH
	}

	if err := unix.Flock(int(f.Fd()), flag); err != nil {
		return errors.Join(ErrLockUnsuccessful, err)
	}
	return nil
}
```

Why not simplify this further and only make doing the actual file lock OS specific but keep the `tryLockFile` with backoff logic the same across OS?

store/posixage/flock.go
joe0BAB: I meant instead of stating in a comment how this should be used (and more importantly where not), you could enforce it by creating a folder `store/posixage/internal/flock` and then in `flock.go` properly export `AttemptLock`. Then, only `store/posixage` would be able to import/use `AttemptLock` and any internal stuff strictly remains internal to flock.go

store/posixage/flock_unix.go
Benehiko: that was what attemptLock was before we changed it and pushed the backoff stuff out of it :sweat_smile:

store/posixage/flock.go
Benehiko: we only want to retry once more if a recovery was possible, not multiple times over.

store/posixage/internal/flock/flock.go
```diff
@@ -0,0 +1,148 @@
...
+	defer lockCtxCancel()
+
+	ep := backoff.NewExponentialBackOff()
+	ep.InitialInterval = time.Millisecond * 10
+	ep.MaxInterval = time.Millisecond * 100
+	_, err := backoff.Retry(lockCtx, func() (bool, error) {
+		if err := lockFile(f.Fd(), exclusive); err != nil {
+			return false, err
+		}
+		return true, nil
```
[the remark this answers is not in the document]
Benehiko: you trying to gaslight me, hmm?

joe0BAB (approved):

PR #294, opened 2025-09-17
feat(daemon): setup OTEL

### LINKED ISSUE #290: OTEL setup
When running the engine daemon as standalone binary, it should be able to pick up / integrate with an existing OTEL setup.
That means two things:
* reading the OTEL endpoint from an ENV
* setting up metric/trace providers in the engine main so that internal OTEL metrics get properly send to that endpoint

joe0BAB:
Implements #290 : 
* `EndpointFromEnv` is a helper function to read the OTEL endpoint ENV
* `InitializeOTel` can be used to wire up OTEL (for now it only has grpc support, http can be added later)
* use both in the standalone engine daemon


x/telemetry/otel.go
```diff
@@ -0,0 +1,154 @@
...
+		k, v, ok := strings.Cut(kv, "=")
+		if !ok {
+			continue
+		}
+		if strings.HasPrefix(k, "OTEL_") && strings.HasSuffix(k, "ENDPOINT") {
+			return v
+		}
+	}
+
+	return ""
```
Benehiko: don't we know these environment variable names beforehand? couldn't we hardcode them to an array and then fetch them from the env using os.Getenv?

x/telemetry/otel.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		otel.Handle(fmt.Errorf("invalid otel endpoint '%s': %s", endpoint, err))
+		return "", false
+	}
+
+	var secure bool
+	switch u.Scheme {
+	// TODO: add support for OTEL endpoints through sockets
+	// case "unix":
+	//	endpoint = unixSocketEndpoint(u)
```
Benehiko, line 129: that's possible?

joe0BAB: yes, eg docker CLI supports it, that's how I know

But not sure if we actually want that too here.

x/telemetry/otel.go
joe0BAB: I think this is a leftover from testing different things locally. It's just one ENV: `OTEL_EXPORTER_OTLP_ENDPOINT`

I'll remove it

Benehiko (approved):

PR #307, opened 2025-10-15
feat(injector): container create rewriter

### LINKED ISSUE #306: secrets in environment variables

joe0BAB:
Implements #306 

Adds a new module `injector` that exports `ContainerCreateRewriter` which is a "rich" secrets engine client that resolves ENVs as specified in #306 into actual secrets.

The code is split into the resolve logic (`resolver`) and the code that maps it into the `container.CreateRequest` as defined from `github.com/docker/docker/api/types/container`. This allows to more easily re-use the resolve logic for other mappers (we'll need a slightly different mapper for NRI later).


injector/injector.go
```diff
@@ -0,0 +1,100 @@
...
+		if err != nil {
+			if !errors.Is(err, secrets.ErrNotFound) {
+				r.logger.Errorf("resolving ENV %s to secret: %s", key, err)
+			}
+			return "", nil
+		}
+		if len(result) == 0 {
+			return "", nil
+		}
+		return string(result[0].Value), nil
```
Benehiko: this is duplicated below

```suggestion
```

Benehiko: we can instead rely on a bool e.g. `isLegacyMode = true` to not return an error etc.

injector/injector.go
```diff
@@ -0,0 +1,100 @@
...
+	return &resolver{
+		logger:   logger,
+		resolver: c,
+	}, nil
+}
+
+func (r *resolver) resolveENV(ctx context.Context, key, value string) (string, error) {
+	if key == "" {
+		return "", errors.New("key is required")
+	}
```
Benehiko: is an empty check necessary? 
Can we just do:

```go
var id secrets.ID
if value != "" {
    // parse value 
} else {
    // parse key
}

```

injector/injector.go
```diff
@@ -0,0 +1,100 @@
...
+			return "", nil
+		}
+		if len(result) == 0 {
+			return "", nil
+		}
+		return string(result[0].Value), nil
+	}
+	if !strings.HasPrefix(value, prefix) {
+		return value, nil
+	}
```
Benehiko: this should be the first line I think.

Operations in order:
1. Check for `value.HasPrefix("se://")`
2. Fallback to legacy mode if `value == ""` parse key
3. Could not do 1 or 2, just return value.

injector/injector.go
```diff
@@ -0,0 +1,100 @@
...
+	return &ContainerCreateRewriter{r: resolver}, nil
+}
+
+func (r *ContainerCreateRewriter) ContainerCreateRequestRewrite(ctx context.Context, req *container.CreateRequest) error {
+	if req.Config == nil {
+		return nil
+	}
+	var resolvedEnvList []string
+	for _, env := range req.Env {
+		key, val, _ := strings.Cut(env, "=")
```
Benehiko, line 107: what happens if the cut failed? e.g. `found = false`.

joe0BAB: it's not failing, the bool value just indicates if `env` contains `=` or not (cf docs)

Benehiko: i meant - _should we fail if it doesn't contain an `=`_?

joe0BAB: it's not on us to validate that here so would just aim to be robust against it. Right now though we would convert a  missing "=" into adding one. I guess maybe we should not do that then?

joe0BAB: I'll change accordingly

injector/injector.go
```diff
@@ -0,0 +1,127 @@
...
+	var err error
+	switch {
+	case value == "":
+		pattern, err = secrets.ParsePattern(key)
+		if err != nil {
+			r.logger.Printf("%s has no value but is not a valid secret ID (%s) -> cannot resolve", key, err)
+			return "", nil
+		}
+	case strings.HasPrefix(value, prefix):
+		pattern, err = secrets.ParsePattern(strings.TrimPrefix(value, prefix))
```
Benehiko: you can _always_ trim the prefix (if no prefix exists, it returns the string unchanged).

Instead we could do an if which sets a variable that can be used in ParsePattern.

e.g.
```go
v := value
if v == "" {
    v = key
}
pattern, err := secrets.ParsePattern(strings.TrimPrefix(v, prefix))
```

injector/injector.go
```diff
@@ -0,0 +1,127 @@
...
+			r.logger.Errorf("resolving ENV %s to secret: %s", key, err)
+		}
+		return "", withNoErrLegacyFallback(err)
+	}
+
+	if len(result) == 0 {
+		return "", withNoErrLegacyFallback(secrets.ErrNotFound)
+	}
+
+	return string(result[0].Value), nil
```
Benehiko, line 86: are we not going to sort by provider alphabetically?

joe0BAB: let's start simple with first match from whatever the engine gives us (right now the engine already sorts plugins alphabetically, that's what I meant when we discussed about it)

injector/injector.go
joe0BAB: we still in addition need an early return but agree, let's do both then as it simplifies readability a lot

Benehiko (approved):
LGTM

PR #309, opened 2025-10-17
refactor(client): do not check for empty path in WithSocketPath client option

### LINKED ISSUE #306: secrets in environment variables

joe0BAB:
Returning an error when creating the client but `WithSocketPath()` has received an empty string is a bit too opinionated / breaks with conventions. 

If that path is empty, it will cause an error later anyway. We should leave it up to the caller when it wants this to be checked.

This allows the caller to write simpler code (eg by guaranteeing a non empty string on compile time already => no need for any runtime error checks => simpler code).


Benehiko (commented):
I think there is still value in returning an error from the `Options` type.

Being opinionated is also fine since we own the client.

PR #311, opened 2025-10-17
feat: export ErrNotFound as ErrSecretNotFound through client/injector/plugin/engine

### LINKED ISSUE #306: secrets in environment variables

joe0BAB:
Another export we missed.


wmluke (approved):
LGTM

PR #338, opened 2025-10-30
engine: move to config interface

Benehiko:
Moving to an interface makes the config way more portable. We get the benefit of expanding the config and only exposing what is needed for downstream components. It's the first step in separating some of the plugin code and engine/server setup code.


engine/engine.go
```diff
@@ -28,6 +30,8 @@ const (
 	engineShutdownTimeout = 2 * time.Second
 )
 
+type Plugin = plugin.Plugin
```
joe0BAB: nit: I'd suggest to move this to `adaptation.go` where we already have the list of exported types at the top

engine/adaptation.go
```diff
@@ -79,34 +75,76 @@ func (c *Config) validated() (metadata, error) {
...
+func (e *engineConfig) PluginPath() string {
+	return e.pluginPath
+}
+
+func (e *engineConfig) Version() string {
+	return e.version
+}
+
+var _ config.Engine = &engineConfig{}
+
```
joe0BAB, line 135: nit: Could we move this into a `config.go`? My idea was to keep the `adaptation.go` limited to contain everything someone importing the engine needs to know, ie mostly exported types and config related stuff (from how you edited the code I'm guessing we should additionally rename this file to `engine.go` and rename `engine.go` to something else)

Benehiko: so this is a first step of a huge refactor, where most files under the `engine` package will move to their own respective packages under `internal`.

engine/adaptation.go
```diff
@@ -79,34 +77,76 @@ func (c *Config) validated() (metadata, error) {
...
 	upCb                   func(ctx context.Context) error
 	tracker                telemetry.Tracker
 }
 
+func (e *engineConfig) Plugins() map[plugin.Metadata]plugin.Plugin {
+	return e.plugins
+}
+
+func (e *engineConfig) PluginsEnabled() bool {
+	return !e.enginePluginsDisabled
```
joe0BAB: `enginePluginsDisabled` is actually `engineLaunchedPluginsDisabled` which is not the same as `PluginsEnabled`

We probably want to rename this to `EngineLaunchedPluginsEnabled`?

Benehiko: yeah i was a bit confused, because the underlying field is called `enginePluginsDisabled`.

engine/engine_test.go
```diff
@@ -199,20 +200,32 @@ func Test_discoverPlugins(t *testing.T) {
 		assert.NoError(t, os.WriteFile(filepath.Join(dir, "text-file"), []byte(""), 0o644))
 		assert.NoError(t, createFakeExecutable(filepath.Join(dir, "binary-file")))
 		assert.NoError(t, createFakeExecutable(filepath.Join(dir, "my-plugin")))
-		plugins, err := scanPluginDir(testhelper.TestLogger(t), dir)
+		cfg := &enginetesthelper.TestEngineConfig{
+			T:              t,
+			TestPluginPath: dir,
+		}
```
joe0BAB, line 206: nit: I'd probably inline these (make it one line I mean)

Benehiko: i don't like that, it makes the line longer than 80 spaces

engine/resolver_test.go
```diff
@@ -74,7 +75,13 @@ func TestResolver(t *testing.T) {
 			newMockResolverRuntime("foo", errors.New("foo")),
 			newMockResolverRuntime("bar", errors.New("bar")),
 		}}
-		resolver := newRegResolver(testhelper.TestLogger(t), tracker, reg)
+		cfg := &enginetesthelper.TestEngineConfig{
+			T:           t,
+			TestTracker: tracker,
+		}
```
joe0BAB, line 81: nit: as in the other file, I'd inline these here too

Benehiko: no

joe0BAB (approved):
LGTM!

PR #339, opened 2025-10-30
engine: routes setup

Benehiko:
This change sets up a common way to add routes and middleware. It improves readability and allows us to extend our routes easily.


engine/internal/routes/routes.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		r.Logger().Printf("registering public route: %s", path)
+		r.registeredPublicRoutes = append(r.registeredPublicRoutes, path)
+
+		// routes that end on the forward-slash '/' indicate that they handle sub-routes
+		// themselves. In an http.ServeMux it would match any route with such a
+		// prefix, but in go-chi/chi we need a wildcard '*'
+		if strings.HasSuffix(string(path), "/") {
+			path = path + "*"
+		}
```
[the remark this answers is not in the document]
Benehiko: readability here is more important for now. Still deciding how to best handle the wildcard.

engine/internal/routes/routes.go
```diff
@@ -0,0 +1,112 @@
...
+	privRoutes = append(privRoutes, r)
+}
+
+type Routes interface {
+	RegisterMiddleware(router *chi.Mux) error
+	RegisterRoutes(router *chi.Mux) error
+}
+
+type routes struct {
+	config.Engine
```
joe0BAB, line 43: Do we need this entire thing? It seems we only use the logger from this, so maybe just have that here?

engine/internal/routes/routes.go
```diff
@@ -0,0 +1,112 @@
...
+	route        func(routeConfig) (routePath, http.Handler, error)
+	publicRoute  route
+	privateRoute route
+)
+
+var errRouteDisabled = errors.New("disabled route")
+
+var (
+	pubRoutes  []publicRoute
+	privRoutes []privateRoute
```
joe0BAB, line 31: Couldn't we move this inside the `routes` struct? Also the way it's used right now is potentially racy.

engine/internal/routes/health.go
```diff
@@ -0,0 +1,16 @@
+package routes
+
+import (
+	"net/http"
+)
+
+func init() {
```
joe0BAB, line 7: This is racy right now. Couldn't we drop this altogether and do it from the program flow?

init is quite a dangerous tool

engine/internal/routes/resolver.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/docker/secrets-engine/x/api/resolver/v1/resolverv1connect"
+)
+
+func resolverHandler(c routeConfig) (routePath, http.Handler, error) {
+	r := service.NewService(c.Logger(), c.Tracker(), c.Registry())
+	path, h := resolverv1connect.NewResolverServiceHandler(resolver.NewResolverHandler(r))
+	return routePath(path), h, nil
+}
+
+func init() {
```
joe0BAB: Same as with the other init.

engine/internal/routes/routes.go
Benehiko: yes we do, check the handlers, they use the `routeConfig` to access logger, tracer, plugin config (soon) and registry.

engine/internal/routes/routes.go
Benehiko: no, `func init` is always called in sequence and first.

engine/internal/routes/health.go
Benehiko: no, it's not. init is called in sequence on startup before `func main()`

engine/internal/routes/resolver.go
Benehiko: no it isn't

joe0BAB: nit: wouldn't we want to move the init to the top though?

engine/internal/routes/routes.go
```diff
@@ -0,0 +1,129 @@
...
+	return r.reg
+}
+
+//
+// func New(cfg config.Engine, reg registry.Registry) Routes {
+// 	return &routes{
+// 		Engine: cfg,
+// 		reg:    reg,
+// 	}
+// }
```
joe0BAB: leftover?

joe0BAB (approved):

PR #345, opened 2025-10-31
engine: move plugin handlers to routes

Benehiko:
Move the plugin handlers to the `routes` package. This required quite a bit of work in defining an interface which the plugin config conforms to. I think we can clean it up more once we have some of these changes moved to the internal package. 

It will take a few iterations I think before everything is neat like we want it. Simplifying how we launch plugins and register them would also allow for less bugs in future, so we should tackle that slowly as well.


engine/plugin.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/docker/secrets-engine/x/ipc"
 	"github.com/docker/secrets-engine/x/logging"
 	"github.com/docker/secrets-engine/x/secrets"
+
+	_ "github.com/docker/secrets-engine/engine/internal/routes/plugin"
```
Benehiko, line 27: This is required for the plugin routes to register themselves to the router. Once we clean up the plugin.go file, we'll have only these imports neatly defined here.

joe0BAB (approved):

PR #358, opened 2025-11-10
feat(pass): make store configurable

joe0BAB:
Service group of the underlaying store in pass should be configurable.


Benehiko (commented):
Maybe we should make this a Variadic option, so we don't have to pass in "" in cases where this doesn't change?

joe0BAB:
But this is docker-pass and not the store library. We always want to specify this here, no?

Benehiko (approved):

PR #359, opened 2025-11-10
plugins/credential-helper: configure custom key rewriter

Benehiko:
Legacy keys inside the credential-helper would sometimes require a prefix for them to be available to certain services. This patch adds the ability to configure a custom rewriter.

Added missing tests by mocking the credential-helper.


plugins/credentialhelper/store.go
```diff
@@ -51,9 +64,14 @@ func (s *credentialHelperStore) GetSecrets(_ context.Context, pattern secrets.Pa
 				s.Warnf("could not get matched secret key '%s' from the credential-helper: %s", serverURL, err)
 				continue
 			}
+			value, err := json.Marshal(cred)
+			if err != nil {
+				s.Errorf("could not marshal credential-helper credential with key '%s'", serverURL)
+				continue
+			}
```
joe0BAB: nit: move into separate PR / commit

joe0BAB:
This PR doesn't do what it says it does. It does multiple more things:
- fixes a bug when reading secrets by marshaling them correctly
- makes the credstore read function more robust
- introduces variadic args
- introduces the rewrite option as in PR / commit description

Could we align things a bit, either by turning this into multiple PRs / adjusting PR description and probably in any case adjust the commits?

joe0BAB (approved):

PR #376, opened 2025-11-25
feat(pass): move prerun wrap into cmd/pass

joe0BAB:
pass: remove dependency to moby/cli by moving the two lines that were wrapping the prerun function up to where the pass command is used


cmd/pass/main.go
```diff
@@ -44,3 +44,16 @@ func main() {
...
 }
+
+func wrapPersistentPreRun(cmd *cobra.Command) *cobra.Command {
+	if plugin.PersistentPreRunE != nil {
+		oldPreRunE := cmd.PersistentPreRunE
+		cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+			if err := oldPreRunE(cmd, args); err != nil {
+				return err
+			}
+			return plugin.PersistentPreRunE(cmd, args)
```
joe0BAB, line 55: I'm not sure if the order of this needs to be reverse

@Benehiko could you double check?

pass/command.go
```diff
@@ -4,7 +4,6 @@ import (
 	"context"
 	"os"
 
-	"github.com/docker/cli/cli-plugins/plugin"
```
joe0BAB, line 7: this is the main problem I'm trying to solve, getting rid of this import

cmd/pass/main.go
Benehiko: I think this order is correct

Benehiko (approved):

PR #382, opened 2025-11-26
feat: module plugins

Benehiko:
This PR establishes a standalone `plugins` module which only exports the interface type used by third-party plugins. This prevents importing the entire engine module and separates some of the plugin logic to the engine logic.


x/plugins/plugins.go
```diff
@@ -0,0 +1,13 @@
...
+	"context"
+
+	"github.com/docker/secrets-engine/x/secrets"
+)
+
+type Plugin interface {
+	secrets.Resolver
+
+	Run(ctx context.Context) error
+}
```
joe0BAB, line 13: why wouldn't we move that into the `plugin` module?

PR #383, opened 2025-11-26
fix: move plugin type to shared x package

Benehiko:
External and Internal plugins should only need to import from the `plugin` module and not need the entire `engine` module.


plugin/plugin.go
```diff
@@ -0,0 +1,5 @@
+package plugin
+
+import "github.com/docker/secrets-engine/x/plugins"
+
+type Plugin = plugins.Plugin
```
joe0BAB, line 5: maybe rename this file so it's more clear that this is only engine/internal plugins? Otherwise it's confusing for people who don't know that they'll start looking here.

joe0BAB (approved):
we can do the renaming later

PR #391, opened 2025-12-03
fix(engine): optimization should check if the request pattern includes a plugin

joe0BAB:


Benehiko (dismissed):

joe0BAB:
lol, I trolled, it's actually a bit more complicated to solve

Benehiko (approved):

PR #404, opened 2026-01-12
feat: new service to list plugins

### LINKED ISSUE #420: Plugin List API
- name
- version
- pattern
- external/internal (bool)

joe0BAB:
Some protobuf changes:
- Renames `EngineService` to `RegisterService`. 
- Adds a new `ListService` that can be used to implement plugin listing (to be implemented in follow up PR)

Broke down the engine service into "RegisterService" and "ListService" to allow for modular and decoupled implementation (plus it also allows for separating the clients and having a dedicated client for the ListService that's different from the RegisterService client).


x/api/resolver/v1/api.proto
```diff
@@ -30,6 +30,33 @@ message RegisterPluginResponse {
...
+  // Kind/category of plugin.
+  PluginType type = 3;
+}
+
+message ListPluginsRequest {}
+
+message ListPluginsResponse {
+  repeated Plugin plugins = 1;
+}
+
```
joe0BAB, line 54: This is the main change.

x/api/resolver/v1/api.proto
```diff
@@ -30,6 +30,33 @@ message RegisterPluginResponse {
   google.protobuf.Duration request_timeout = 3;
 }
 
+service ListService {
+  // ListPlugins returns all plugins registered with the engine.
+  rpc ListPlugins(ListPluginsRequest) returns (ListPluginsResponse);
+}
+
+enum PluginType {
+  PLUGIN_TYPE_UNSPECIFIED = 0;
```
joe0BAB: buf lint wanted me to add this

x/api/resolver/v1/api.proto
```diff
@@ -30,6 +30,35 @@ message RegisterPluginResponse {
...
+message Plugin {
+  // Name
+  string name = 1;
+  // Version
+  string version = 2;
+  // Pattern
+  string pattern = 3;
+  // Type
+  PluginType type = 4;
+}
```
joe0BAB, line 47: This essentially corresponds to 
```go
type Metadata interface {
	Name() api.Name
	Version() api.Version
	Pattern() secrets.Pattern
}
```
except that we also need to add the plugin type on the protobuf.

Benehiko (approved):

x/api/resolver/v1/api.proto
```diff
@@ -30,6 +30,35 @@ message RegisterPluginResponse {
...
+service ListService {
+  // ListPlugins returns all plugins registered with the engine.
+  rpc ListPlugins(ListPluginsRequest) returns (ListPluginsResponse);
+}
+
+enum PluginType {
+  PLUGIN_TYPE_UNSPECIFIED = 0;
+  PLUGIN_TYPE_BUILTIN = 1;
+  PLUGIN_TYPE_MANAGED = 2;
+  PLUGIN_TYPE_EXTERNAL = 3;
```
Benehiko: Yeah i see here we are doing that as well. We should stick with 2 terms (internal and external). No need to have more than 2 plugin types.

Benehiko (changes requested):

x/api/resolver/v1/api.proto
joe0BAB: I replaced with a boolean `external`

Benehiko (approved):

PR #406, opened 2026-01-13
feat(engine): add runtime type information to runtime interface

joe0BAB:
Extends the runtime interface to provide type information. 

Notes on internal implementation details: This is implemented as string vs using reflections as eg externally and internally launched plugins share the same go struct / type but differ only in how it gets initialized.


engine/internal/plugin/plugin.go
```diff
@@ -23,6 +31,8 @@ type Runtime interface {
 
 	Metadata
 
+	Type() Type
```
joe0BAB, line 34: This is the main change.

joe0BAB:
solved this differently

PR #407, opened 2026-01-13
refactor(engine): move builtin runtime implementation into internal directory

joe0BAB:
Have to move some code around to avoid cyclic imports when implementing the `ListService`.


joe0BAB:
I'll try to add the external runtime code into the same directory as well in a follow up PR.

Benehiko (approved):

PR #409, opened 2026-01-14
feat(engine): list service

### LINKED ISSUE #420: Plugin List API

joe0BAB:
Implements the list service that allows listing plugins currently registered to the engine.

(I'll add more tests for this as part of the existing test suits once we are aligned on how to setup a client for this => follow up PR)


engine/internal/runtime/builtin/builtin.go
```diff
@@ -111,3 +111,8 @@ func (i *internalRuntime) Close() error {
 func (i *internalRuntime) Closed() <-chan struct{} {
 	return i.closed
 }
+
+func IsBuiltin(r plugin.Runtime) bool {
+	_, ok := r.(*internalRuntime)
+	return ok
+}
```
joe0BAB, line 118: I had cyclic import issues as runtime package imports the routes package so moved it into a nested package..

Benehiko (approved):

engine/internal/routes/list.go
```diff
@@ -0,0 +1,45 @@
...
+	"google.golang.org/protobuf/proto"
+
+	"github.com/docker/secrets-engine/engine/internal/registry"
+	"github.com/docker/secrets-engine/engine/internal/runtime/builtin"
+	resolverv1 "github.com/docker/secrets-engine/x/api/resolver/v1"
+	"github.com/docker/secrets-engine/x/api/resolver/v1/resolverv1connect"
+)
+
+func init() {
+	registerPrivateRoute(listHandler)
```
Benehiko, line 17: maybe this should be a public route?

joe0BAB: right, I'll do in a follow up PR

PR #412, opened 2026-01-15
feat: add gomodguard check

joe0BAB:
The main go sub modules should not depend on each other. However, this only is true if enforced via a linter. This PR addresses this and sets up a new linter CI pipeline.

It's inspired by the discussions around https://github.com/etcd-io/etcd/pull/20748 and the current limitations of golang-lint-ci and depguard.


engine/.gomodguard.yaml
```diff
@@ -0,0 +1,13 @@
+---
```
joe0BAB, line 1: Adding the linter revealed that this module in fact does cross reference other internal modules still.

injector/.gomodguard.yaml
Benehiko, line 1: I guess the injector also shouldn't import the engine module?

joe0BAB: yes, once we have the e2e module. injector has a test dependency to engine.

Benehiko (approved):

PR #418, opened 2026-01-16
feat: injector should not link engine or plugin module

joe0BAB:
Replaces the engine dependency with a simple mock engine in the injector unit test => decouples the injector module from the engine.


Benehiko (commented):
I'd rather short-circuit it at the client (resolver). Implement the Resolver interface so that `GetSecrets` would return the mock values instead.
https://github.com/docker/secrets-engine/blob/29cb462a567593f67ba88ee5e06aaa578db62c87/injector/injector.go#L48

Reasons:
- no socket path needed
- no listener needed

injector/injector.go
```diff
@@ -76,7 +76,7 @@ func (r *resolver) resolveENV(ctx context.Context, key, value string) (string, e
 	}
 	getSecrets := func(ctx context.Context, pattern secrets.Pattern, src string) ([]secrets.Envelope, error) {
 		result, err := r.resolver.GetSecrets(ctx, pattern)
-		if err == nil {
+		if err == nil && len(result) > 0 {
```
joe0BAB, line 79: this actually was a tiny bug

Benehiko (approved):
So much better!

PR #419, opened 2026-01-16
feat(client): list plugins method

### LINKED ISSUE #420: Plugin List API

joe0BAB:


client/client_test.go
```diff
@@ -0,0 +1,144 @@
...
+			Pattern:  proto.String(pattern),
+			External: proto.Bool(plugin.External),
+		}.Build())
+	}
+	return connect.NewResponse(resolverv1.ListPluginsResponse_builder{
+		Plugins: plugins,
+	}.Build()), nil
+}
+
+func mockEngine(t *testing.T, socketPath string, list []PluginInfo) {
```
Benehiko: this seems specific to the list plugins. Either we rename it to be specific to plugins, or we make a struct as the "opts" to add more data (thinking of when we have more endpoints later).

client/client_test.go
```diff
@@ -0,0 +1,144 @@
...
+		shutdownCtx, cancel := context.WithTimeout(t.Context(), 5*time.Second)
+		defer cancel()
+
+		_ = server.Shutdown(shutdownCtx)
+		_ = listener.Close()
+		_ = os.RemoveAll(socketPath)
+	})
+}
+
+func testEngine(t *testing.T, plugins []PluginInfo) string {
```
Benehiko: why not merge testEngine with mockEngine?

client/client_test.go
joe0BAB: fair, adjusted accordingly

Benehiko (approved):

PR #440, opened 2026-01-29
feat: (generic realm) expansion

joe0BAB:
We want a generic logic that allows to "resolve a realm".

Example: Let `docker/auth/**` be a realm. Then, the pattern query `**/foo` should get resolved to `docker/auth/**/foo`.

We can define `**` to be a generic realm. Then this reframes our problem into a "generic realm replacement problem".

As there could be `n` generic realms, more generally speaking we can view this a "replace n" problem where we have a component string and up to `n` realm markers `**` that should be replaced.


x/secrets/pattern.go
```diff
@@ -134,3 +155,35 @@ func Filter(filter, other Pattern) (Pattern, bool) {
...
+		}
+	}
+	if len(candidates) != len(other) {
+		return "", fmt.Errorf("%d dimensional realm cannot be resolved via %d pattern(s)", len(candidates), len(other))
+	}
+	for idx, candidate := range candidates {
+		components[candidate] = other[idx]
+	}
+	return strings.Join(components, "/"), nil
+}
```
joe0BAB: this is just for demonstrating the n dimensional case, will delete before merging

x/secrets/pattern.go
```diff
@@ -134,3 +154,18 @@ func Filter(filter, other Pattern) (Pattern, bool) {
...
+func replace1(original, other string) (string, error) {
+	components := split(original)
+	var candidates []int
+	for idx, val := range components {
+		if val == "**" {
+			candidates = append(candidates, idx)
+		}
+	}
+	if len(candidates) != 1 {
+		return "", errors.New("invalid")
```
joe0BAB: @Benehiko any better ideas here?

x/secrets/pattern.go
```diff
@@ -134,3 +154,18 @@ func Filter(filter, other Pattern) (Pattern, bool) {
...
+	}
+	components[candidates[0]] = other
```
Benehiko, line 170: now that we only have 1 replace wouldn't this end up like:

```
pattern: docker/**/bar/**
input ID: bob
output: docker/bob/bar/**
```

joe0BAB: So we could implement what you're describing, but right now it isn't. Instead if the candidate positions where we could replace are not equal 1, we return an error.

I'm hesitant to do this as eg the following could be a bit surprising:
```
pattern: docker/**/bar/**
input Pattern: bob/**
output: docker/bob/**/bar/**
```
🤷

joe0BAB: It also complicates things as with 
```
pattern: docker/**/bar/**
input ID: bob
output: docker/bob/bar/**
```
you're converting an ID into a pattern. Then it wouldn't be `ExpandID(ID) (ID, error)` anymore.

Benehiko: yes, i was asking if the output _would be_ not _should be_.

x/secrets/pattern.go
Benehiko: maybe something like this

```suggestion
		return "", fmt.Errorf("expand only supports one expansion glob, pattern %s has zero or many", original)
```

Benehiko (approved):

PR #442, opened 2026-02-04
feat: govulncheck

joe0BAB:
Adds `govulncheck` CI step and fixes all currently detected issues (bumping Go version was enough to fix them).


Dockerfile
```diff
@@ -77,6 +77,28 @@ RUN --mount=type=bind,target=.,ro \
...
+    go install "golang.org/x/vuln/cmd/govulncheck@latest" \
+    && govulncheck -version
+
+FROM golang AS do-govulncheck
+ARG TARGETARCH
+ARG GO_VERSION
+COPY --link --from=govulncheck /go/bin/govulncheck /go/bin/govulncheck
+WORKDIR /govulncheck
+ENV GOARCH=${TARGETARCH}
+ENV GOTOOLCHAIN=go${GO_VERSION}
```
joe0BAB, line 93: @Benehiko I don't fully get why this is needed. But somehow when I do `go version` it's not the same as `GO_VERSION`. As a result govulncheck would run with the wrong version unless I explicitly set it.

Any ideas? Looks like a potential issue with `FROM golang` but no clue what specifically.

Benehiko: I think we are setting the go version in a couple of places:
- https://github.com/docker/secrets-engine/blob/main/Dockerfile#L3
- https://github.com/docker/secrets-engine/blob/main/go.work#L1
- https://github.com/docker/secrets-engine/blob/main/Makefile#L21

Might be that the environment variable ends up being different from the pulled image (defaults to latest).

joe0BAB: I just don't get how they can divert, cause if I `echo $GO_VERSION` it inside the govulcheck stage it is the right one from the go.work file, ie passing it through the Makefile works as expected. So don't get how in the same Dockerfile it can be wrong / default to the default=latest when we create the gobase stage but it's correct in the later stage?

Benehiko: no idea 🙈

Benehiko (approved):

PR #461, opened 2026-02-18
feat: check dial errors and return appropriate error

Benehiko:
This patch adds a new error `ErrSecretsEngineNotAvailable` which is returned whenever there is a dial error (i.e. secrets engine not reachable).


client/client.go
```diff
@@ -130,6 +138,27 @@ type Client interface {
 	ListPlugins(ctx context.Context) ([]PluginInfo, error)
 }
 
+func isDialError(err error) bool {
+	if err == nil {
+		return false
+	}
+	var oe *net.OpError
+	if errors.As(err, &oe) && (oe.Op == "dial" || oe.Op == "connect") {
```
joe0BAB, line 147: Does `errors.Is` not work here?

client/client.go
```diff
@@ -130,6 +138,27 @@ type Client interface {
 	ListPlugins(ctx context.Context) ([]PluginInfo, error)
 }
 
+func isDialError(err error) bool {
```
joe0BAB, line 142: Couldn't we write some unit tests for this? Eg dial against a non existing socket, dial against a socket that is just a file. Just set the timeouts very short.

Benehiko: added

joe0BAB (dismissed):

client/client.go
Benehiko: it's a struct, so no.

joe0BAB (approved):

PR #474, opened 2026-02-27
fix(keychain): store large blobs on windows

Benehiko:
Some credentials - especially auth tokens, can be quite large. On windows the wincred store only supports up to 2560 bytes in length. This solution below chunks the credential into parts and stores those parts alongside the main ID. 

Chunked credentials are always skipped when filtering credentials from the store by checking the `chunkCountKey` from the "main" credentials metadata. 

The `safelyCleanMetada` function removes the attribute when reading the credential from the store.


store/keychain/keychain_windows.go
```diff
@@ -246,19 +370,36 @@ func (k *keychainStore[T]) Filter(ctx context.Context, pattern store.Pattern) (m
...
+			count, err := strconv.Atoi(countStr)
+			if err != nil {
+				return nil, fmt.Errorf("invalid chunk count %q: %w", countStr, err)
+			}
+			rawBlob, err = k.readChunks(id, count)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			rawBlob = gc.CredentialBlob
```
joe0BAB, line 387: would it make sense to dedupe this? Seeing the same code above.

store/keychain/keychain_windows_test.go
```diff
@@ -25,6 +25,73 @@ import (
...
+			{Keyword: serviceNameKey, Value: []byte("name")},
+		}
+		assert.False(t, isChunkCredential(attrs))
+	})
+	t.Run("returns false for empty attributes", func(t *testing.T) {
+		assert.False(t, isChunkCredential(nil))
+		assert.False(t, isChunkCredential([]wincred.CredentialAttribute{}))
+	})
+}
+
```
joe0BAB, line 94: Don't we want some more tests that verify Save/Filter/Get/Delete on credentials that need to be chunked work as expected?

store/keychain/keychain_windows.go
Benehiko: verbosity is best here

joe0BAB (approved):
LGTM, tweaks are non blocking, let's do them in a follow up

store/keychain/keychain_windows_test.go
Benehiko: definitely, but i think this is good enough to get this fix in quickly

PR #493, opened 2026-03-17
fix: prevent panic when home directory cannot be determined

pulkit-999:
Currently, getConfigPath() can return an empty string if the home directory cannot be determined (e.g., in minimal containers or
misconfigured systems). This empty string is then passed to path.Join(), causing a panic: "runtime error: makeslice: len out of range".

This change makes getConfigPath() return an error instead, and adds proper error handling in the caller. The error is wrapped with context to help users understand what went wrong.


Benehiko:
If the home directory is misconfigured and the software cannot determine it anyway, why would a panic be the incorrect behavior?

pulkit-999:
Missing home directories can happen in minimal containers, CI environments, or misconfigured systems. Expected runtime conditions should return errors, not panic. This provides better user experience.

Benehiko:
Can you provide a PoC? show me what errors you see (i.e. what the panic does). 
The code above is basically the same as the docker/cli and it runs inside containers and CI as well.

PR #501, opened 2026-03-25
feat: writer endpoint

Benehiko:


Benehiko:
Closing this draft — it needs to be reworked.

This SDK-only sketch predates the design refinement on the write endpoint and doesn't address the core security considerations (caller authorization, per-key scoping, default-store-only enforcement). The endpoint path, overwrite semantics, and error taxonomy were also chosen before the threat model was settled.

Tracking the refined requirements in an internal issue. A reworked implementation should follow once that issue is settled.

PR #502, opened 2026-03-25
feat(api): add daemon version endpoint client and version comparison

Benehiko:
Add GetDaemonVersion to query the /version endpoint exposed by the engine, returning a DaemonVersion (name + version) or ErrVersionNotSupported when the daemon predates the version API.

Add Compare(a, b Version) for semver ordering, enabling callers to gate feature use on the daemon version.


x/api/daemon.go
```diff
@@ -0,0 +1,21 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+// DaemonVersion holds the name and version reported by the daemon.
+type DaemonVersion struct {
+	Name    string
```
joe0BAB: Could we use the `Name` type here maybe?

joe0BAB: see the `name.go`

client/client.go
```diff
@@ -118,8 +121,10 @@ type config struct {
 }
 
 type client struct {
+	httpClient     *http.Client
```
joe0BAB: Why do we need this one? Maybe I'm missing something, but I don't see it being used 🤔

joe0BAB (approved):

PR #512, opened 2026-04-23
feat(pass): documentation

joe0BAB:


plugins/pass/command.go
```diff
@@ -47,10 +48,74 @@ Examples:
...
+
+` + "```" + `console
+$ docker pass set GH_TOKEN=123456789
+` + "```" + `
+
+Create a secret from STDIN:
+
+` + "```" + `console
+echo 123456789 > token.txt
+cat token.txt | docker pass set GH_TOKEN
```
Benehiko: let's maybe not show from a file like this?

We should show piping the value instead? and what about no value and just getting prompted?

Benehiko (approved):

PR #523, opened 2026-05-21
posixage: use caller context for lock waits

### LINKED ISSUE #525: Simplify posixage lock acquisition API
### Problem

  The shape of the API conflates two unrelated concepts:

  1. **`LockTimeout` is a wait duration** — exactly what `context.Context`
     exists for. Callers already pass a `ctx` to `Save`, `Get`, etc. Adding
     `WithLockTimeout` on top means callers now have two ways to express
     "how long am I willing to wait" — one via the ctx they pass per-call,
     and one via a store-level option configured at construction. This
     becomes ambiguous when both are set (the implementation picks
     whichever fires first), and it forces callers to think about lock
     waits separately from operation deadlines.

  2. **`StaleLockTimeout` is a file-age threshold**, not a wait duration —
     it controls when a `.posixage.lock` file is considered abandoned and
     may be removed. It's a semantically distinct concept that genuinely
     belongs as a store-level option.

  Bundling both into a `flock.Config` struct passed to every `TryLock` call
  obscures this distinction and adds API surface (a struct, two options,
  two new validation paths) for capability that is already expressible via
  `context.Context`.

  ### Proposal

  Drop `WithLockTimeout` and the `flock.Config` struct entirely. Lock
  acquisition retries until the caller context is canceled — callers that
  want a bounded wait pass `context.WithTimeout(ctx, …)`; callers that want
  no timeout (the disable-timeout use case the PR was motivated by) pass a
  context without a deadline.

  Keep `WithStaleLockTimeout` as the only locking knob, since it controls
  a file-age threshold rather than a wait duration. The single-attempt
  stale-recovery trigger moves from "after `LockTimeout` expires" to "after
  the first failed acquisition attempt" — semantically the same, but no
  longer dependent on a separately-configured wait duration.

  ### Resulting API surface

  ```go
  // flock package
  func TryLock(ctx context.Context, root *os.Root, staleAfter time.Duration) (UnlockFunc, error)
  func TryRLock(ctx context.Context, root *os.Root, staleAfter time.Duration) (UnlockFunc, error)

  // posixage store
  posixage.WithStaleLockTimeout(d time.Duration) Options  // unchanged in spirit
  // posixage.WithLockTimeout — removed

  Trade-off

  Callers lose the ability to set a short lock-retry while keeping a long
  overall operation deadline (e.g. "this Save can run for 30s but fail
  the lock acquisition after 100ms"). This isn't a real use case in
  practice: if the lock is held by a live process for longer than 100ms,
  stale recovery won't help (the file isn't actually stale), so the caller
  would just fail. The previous 100ms default was effectively a fail-fast
  heuristic, which the new first-attempt-then-recover flow expresses more
  directly.

  Why this matters now

  Doing this refactor before #521 merges (or as a follow-up before it
  ships externally) avoids landing an API we already know we want to
  remove. WithLockTimeout would become a deprecation-and-removal cycle
  otherwise.

ilopezluna:
- use the caller context to bound lock acquisition instead of the previous internal 100ms wait
- keep stale lock recovery internal with the existing fixed 30s age threshold
- update posixage flock tests and README locking docs for context-driven waits

Verification:
- `go test ./store/posixage/...`
- `git diff --check`


store/posixage/internal/flock/flock.go
```diff
@@ -30,14 +30,38 @@ var (
...
+}
+
+// DefaultConfig returns the lock configuration used when callers do not
+// override lock behavior.
+func DefaultConfig() Config {
+	return Config{
+		LockTimeout:      defaultLockTimeout,
+		StaleLockTimeout: defaultStaleLockTimeout,
+	}
+}
```
Benehiko: make this a var, no need to have a function

Benehiko (changes requested):
The capability is right, but the API shape needs rework before merging.

 Drop WithLockTimeout and flock.Config.LockTimeout. context.Context already expresses "how long to wait" — callers who want a bounded wait pass context.WithTimeout; callers who want no timeout pass a ctx without a deadline. That gives you the disable-timeout case for free, no new option needed.

 Keep WithStaleLockTimeout only. It's a file-age threshold, not a wait duration — semantically distinct, belongs as a store-level option.

 Move stale recovery trigger off LockTimeout. With LockTimeout gone, trigger recovery on first-attempt failure instead (try once → on fail, attempt recovery if StaleLockTimeout > 0 and ctx not canceled → retry until ctx done).

 Proposed signature: 
 ```go
 func TryLock(ctx context.Context, root *os.Root, staleAfter time.Duration) (UnlockFunc, error)
```

Defaults stay (no deadline = wait forever; 30s stale). Drops one option, one struct, and the LockTimeout/StaleLockTimeout coupling.

store/posixage/internal/flock/flock.go:
  - tryLock — drop cfg Config param, take staleAfter time.Duration; restructure to "first-attempt → recover → retry til ctx done"
  - retryLock — drop timeout param, retry purely against ctx
  - TryLock / TryRLock — new signatures (ctx, root, staleAfter)
  - Delete Config, DefaultConfig, defaultLockTimeout const

  store/posixage/store.go:
  - fileStore.tryLock / tryRLock — pass f.staleLockTimeout instead of f.lockConfig
  - config struct — replace lockConfig flock.Config with staleLockTimeout time.Duration
  - Delete WithLockTimeout
  - WithStaleLockTimeout — stays, writes to new field
  - New — init default staleLockTimeout = 30 * time.Second (move const from flock pkg or expose getter)

  store/posixage/internal/flock/flock_test.go:
  - All tests passing Config{...} — rewrite to pass staleAfter directly; tests covering LockTimeout behavior become ctx-deadline
  tests

  store/posixage/store_test.go:
  - Any test exercising WithLockTimeout — remove or convert to ctx-deadline

  store/posixage/README.md:
  - Update the new lock-config section

Benehiko (approved):

PR #542, opened 2026-06-01
fix(keychain): close D-Bus connection after each Linux operation

### LINKED ISSUE #541: Unclosed DBus connections lead to leaked goroutines and open file descriptors
1. Clone this repository.
2. Create `main.go` with the following content in a new folder named `./store/keychain/cmd/dbusleak/`.
3. Run the test program with `dbus-run-session`.
```
$ dbus-run-session -- go run ./store/keychain/cmd/dbusleak -n 200
before  200 calls: goroutines=  1 open-fds=  5
after   200 calls: goroutines=401 open-fds=207
```
```Golang
//go:build linux

// Command dbusleak reproduces the D-Bus connection leak in the Linux keychain
// store.
//
// Every keychain operation (Get/Save/Delete/Filter/GetAllMetadata in
// keychain_linux.go) starts by calling secretservice.NewService(), which opens
// a private session-bus connection via dbus.ConnectSessionBus(). That
// connection is never closed: SecretService has CloseSession() (which only
// closes the remote Secret Service *session*) but no Close() for the underlying
// *dbus.Conn. Each leaked connection keeps a socket FD open plus two goroutines
// (a reader and a signal handler) alive until the process exits.
//
// The leak happens at connect time, before any Secret Service call, so this
// needs only a session bus -- no gnome-keyring/KWallet required:
//
//	dbus-run-session -- go run ./store/keychain/cmd/dbusleak -n 200
//
// Expected output against the current code: goroutines grow by ~2*N and open
// FDs by ~N. Once SecretService gains a Close() that the keychain_linux.go
// callers defer, both counts stay flat.
package main

import (
	"flag"
	"fmt"
	"os"
	"path/filepath"
	"runtime"

	"github.com/docker/secrets-engine/store/keychain/internal/go-keychain/secretservice"
)

func openFDs() int {
	entries, _ := os.ReadDir(filepath.Join("/proc", fmt.Sprint(os.Getpid()), "fd"))
	return len(entries)
}

func main() {
	n := flag.Int("n", 200, "number of NewService() calls")
	flag.Parse()

	fmt.Printf("before %4d calls: goroutines=%3d open-fds=%3d\n", *n, runtime.NumGoroutine(), openFDs())

	for i := 0; i < *n; i++ {
		// Exactly what every keychainStore method does first. The returned
		// *SecretService owns a *dbus.Conn that nothing ever closes.
		if _, err := secretservice.NewService(); err != nil {
			fmt.Fprintln(os.Stderr, "NewService:", err)
			os.Exit(1)
		}
	}

	runtime.GC() // prove the leak survives GC: the blocked goroutines pin everything
	fmt.Printf("after  %4d calls: goroutines=%3d open-fds=%3d\n", *n, runtime.NumGoroutine(), openFDs())
}
```

eginez:
## Summary

On Linux, every keychain operation leaked one socket file descriptor. Long-lived
processes (e.g. a daemon that reads credentials periodically) accumulated these
until they hit the session bus's `max_connections_per_user` limit and could no
longer open any connection, failing with *"The maximum number of active
connections has been reached"*. This closes the connection after each operation.

## What changed

- Added `SecretService.Close()` in `store/keychain/internal/go-keychain/secretservice/secretservice.go`, which closes the underlying `*dbus.Conn`. It's nil-guarded, idempotent (godbus closes the connection once), and tearing down the connection also stops the signal goroutine that `NewService` starts.
- Deferred `service.Close()` immediately after `NewService()` in all five Linux ops — `Get`, `Save`, `Delete`, `GetAllMetadata`, `Filter` (`store/keychain/keychain_linux.go`). It's deferred *before* `CloseSession`, so by LIFO the session (which makes a live D-Bus call) closes first and the connection last. This also covers a path that previously leaked when `OpenSession` failed after `NewService` succeeded.
- Added regression test `TestKeychainDoesNotLeakConnections` (`store/keychain/keychain_linux_test.go`, `//go:build linux`).

## Why

Each Linux op opens a fresh private session-bus connection via `NewService()` →
`dbus.ConnectSessionBus()`, but the code only deferred `CloseSession(session)`,
which releases the secret-service *session* object — not the underlying
`*dbus.Conn`. The connection (and its fd) stayed open for the lifetime of the
process, one per operation.

Scope is Linux-only. The darwin and windows backends use native OS APIs (no
D-Bus) and were never affected.

## Testing

- `DOCKER_TARGET=ubuntu-24-gnome-keyring make keychain-linux-ci-unit-tests` — pass.
- `DOCKER_TARGET=fedora-43-gnome-keyring make keychain-linux-ci-unit-tests` — pass.
- The new test counts open fds in `/proc/self/fd` across 30 lookups of a
  non-existent id. That id drives the full connection setup (`NewService` →
  `OpenSession` → resolve collection → search) but returns `ErrCredentialNotFound`
  before fetching a secret, so it exercises the connection lifecycle without
  mutating shared keyring state. It asserts the open-fd count stays flat.
- Reverting only the fix makes the new test fail with *"open fd count grew by 30
  over 30 lookups"*, confirming a true red-before / green-after regression test.


store/keychain/keychain_linux_test.go
```diff
@@ -17,13 +17,74 @@
...
 )
 
+// openFDCount returns the number of open file descriptors held by the current
+// process by counting entries in /proc/self/fd (Linux only).
+func openFDCount(t *testing.T) int {
+	t.Helper()
+	entries, err := os.ReadDir("/proc/self/fd")
+	require.NoError(t, err)
+	return len(entries)
+}
```
Benehiko, line 37: Let's drop this test. I don't like depending on assertions that depend on the OS environment staying a certain way. We most definitely will see flakes on this.

store/keychain/keychain_linux_test.go
```diff
@@ -17,13 +17,74 @@
...
+// the process's open-fd count stays bounded instead of growing
+// one-per-operation.
+//
+// The lookups target a non-existent id so the test exercises the full
+// connection setup (NewService -> OpenSession -> resolve collection ->
+// SearchCollection) and then returns ErrCredentialNotFound *before* fetching a
+// secret. That keeps the test focused on the connection lifecycle — the thing
+// the fix changes — without creating, reading, or deleting shared keyring
+// items (which is both stateful and prone to gnome-keyring item-lock quirks).
+func TestKeychainDoesNotLeakConnections(t *testing.T) {
```
Benehiko, line 56: Drop these tests - I have a follow-up PR which will introduce mock service which at least helps against regressions https://github.com/docker/secrets-engine/pull/543

PR #555, opened 2026-06-11
chore(docs): avoid saving secret to .bash_history and showing it on the screen

nikelborm:
## Why?

People tend to copy-paste the code from the README


nikelborm:
~Although it should be noted that this approach is not ideal either, because the password will still be temporarily visible in, for example, htop, when the command line is expanded~

UPD. Okay, I looked through the source, and it actually supports reading from stdin already without the need for any bash snippet (`read -s -p "Enter the secret:" secret && docker pass set foo=$secret`).

I've updated the PR to show people 2 available ways to set the secret, and the consequences of choosing the way that was shown initially

Benehiko (approved):

PR #556, opened 2026-06-11
refactor(pass): pull store from cobra context

joe0BAB:
Subcommands previously took a fully-initialized store.Store as a constructor argument, which forced the caller to open the OS keychain before cobra had even dispatched the subcommand. That eager init is the root cause of the headless-hang reported in #405: on systems with no D-Bus session bus the keychain dial blocks indefinitely, taking down commands that never needed the keychain in the first place.

Drop the store argument from GetCommand/SetCommand/ListCommand/RmCommand and have each RunE pull the live store from the cobra command context via the new commands.StoreFrom helper. ErrNoStoreInContext is returned when no store has been attached, so a subcommand that runs without the root having opened the keychain fails with a clear error rather than dereferencing nil.

Root and the OTel wrapper live in the consuming binary (secrets-engine- private) — once it switches to a PersistentPreRunE that calls commands.WithStore lazily, the headless `pass version` / `--help` paths will no longer touch the keychain.

Tests now exercise each subcommand directly, attaching a mock store via WithStore and running the command in isolation. The telemetry test was tied to the now-removed withOTEL wrapper and has been dropped along with the otel/term direct dependencies; vendor/ is updated accordingly.


joe0BAB:
Replaces https://github.com/docker/secrets-engine/pull/554

Benehiko (approved):

PR #557, opened 2026-06-19
fix(keychain): update Secret Service items in place on Linux to stop duplicate accumulation

Benehiko:
## What

On Linux, `keychainStore.Save` copied the secret's volatile `Metadata()` into the **searchable** Secret Service attributes and relied on `CreateItem(ReplaceBehaviorReplace)` to overwrite the prior item. Both gnome-keyring and kwalletd select the replace target by matching the **full** supplied attribute set, so any change in the volatile metadata (e.g. the Docker Hub OAuth credential's rotating JWT claims) defeats the match and a **brand-new item is created on every save** — duplicates pile up without bound, and each stale item keeps a cleartext copy of the old claims.

Fixes the write side of an internally-tracked report and https://github.com/docker/sbx-releases/issues/245.

## How

`Save` now updates **in place** instead of relying on the daemon's replace-match:

1. Search by the stable identity triple `{service:group, service:name, id}` **only** — never the volatile metadata — so a changed metadata value can never hide a previously-stored item.
2. **Absent** → `CreateItem`.
3. **Present** → rewrite the first match's secret (`SetItemSecret`), then best-effort refresh its attributes/label and delete the remaining duplicates.

The item's object path is preserved, so the secret is never momentarily absent and no duplicate is minted. To enable the in-place update, three thin `org.freedesktop.Secret.Item` wrappers were added to the vendored `secretservice` library: `SetItemSecret`, `SetItemAttributes`, `SetItemLabel`.

This is **backend-agnostic** — the attribute-match behaviour is shared by gnome-keyring and kwalletd, verified against both daemons' source. macOS and Windows key items on a stable identifier (`Account` / `TargetName`) and are unaffected.

## Contract

Unchanged: `Save` still returns `nil` iff the secret is stored. Refreshing attributes/label and collapsing leftover duplicates are best-effort and never flip the result. `Delete` is untouched.

## Tests

- **Unit (fake-backed):** `TestKeychainSaveCreatesWhenAbsent` (create branch), `TestKeychainSaveCollapsesDuplicatesInPlace` (in-place update + collapse).
- **Integration (real gnome-keyring harness):**
  - `TestKeychainCollapsesExistingDuplicates` — seeds a 3-item duplicate backlog directly via the daemon, then asserts a single `Save` collapses it to one item holding the latest secret.
  - `TestKeychainSaveDoesNotAccumulate` — five saves with changing metadata stay at exactly one item, with the surviving item's metadata refreshed in place.

All pass against real gnome-keyring; lint clean (`golangci-lint v2.12.2`). The kwalletd target is daemon-agnostic but remains disabled in the harness (prompts under headless).


store/keychain/keychain_linux.go
```diff
@@ -385,13 +388,43 @@ func (k *keychainStore[T]) Save(_ context.Context, id store.ID, secret store.Sec
...
+	// best-effort (the secret is already stored) and must not flip the result.
+	primary := items[0]
+	if err := service.SetItemSecret(primary, sessSecret); err != nil {
+		return err
+	}
+	_ = service.SetItemAttributes(primary, attributes)
+	_ = service.SetItemLabel(primary, label)
+	for _, dup := range items[1:] {
+		_ = service.DeleteItem(dup)
+	}
```
Benehiko, line 513: migration step for secrets that are currently being duplicated. future writes won't duplicate

joe0BAB: Shall we put a comment that this can be deleted from the code eventually?

joe0BAB (dismissed):

joe0BAB (approved):

PR #560, opened 2026-06-24
fix(keychain): retry collection ops when gnome-keyring relocks

Benehiko:
The Linux keychain store dials a fresh D-Bus connection per operation and closes it on return. gnome-keyring scopes an unlock to the session that performed it, so closing a previous connection relocks the collection. That relock can land asynchronously mid-operation — after IsLocked reports the collection unlocked but before the call against the collection runs — surfacing an intermittent "Cannot create an item in a locked collection" error.

Add withRelockRetry, which reacts to the authoritative signal (the operation's own org.freedesktop.Secret.Error.IsLocked D-Bus error) by unlocking and retrying with bounded exponential backoff. Wrap the mutating and reading calls (CreateItem, DeleteItem, GetSecret) with it, and factor the unlock preamble into ensureUnlocked. A failed Unlock (e.g. dismissed auth prompt) aborts the loop immediately to avoid re-prompting.


Benehiko:
Superseded by #557, which merged the relock-retry mechanism (`withRelockRetry` / `isLockedDBusError`) and wrapped the write-path mutations (Save's CreateItem/SetItemSecret/duplicate-collapse DeleteItem, and Delete). That makes this PR's helper, its Delete wrap, and its now-obsolete single-CreateItem Save change redundant/conflicting against main.

The one piece not yet on main is the **read-path** relock retry (Get and Filter/loadSecret's GetSecret). Opening a focused follow-up PR for just that, on top of the merged helper.

Benehiko:
Follow-up with the read-path relock retry: #561

PR #564, opened 2026-06-25
refactor(api): use abstract Unix socket for daemon on Linux

joe0BAB:
Split the non-Windows DaemonSocketPath into a darwin-only filesystem path and a Linux implementation that returns an abstract Unix domain socket address (leading "@", which Go maps to a NUL byte). Abstract sockets live in the abstract namespace rather than on the filesystem, so there are no stale socket files, no directory permission concerns, and the socket is reclaimed automatically when the daemon exits. The address is namespaced by UID to avoid collisions between users.

The plugin config test skips filesystem setup for abstract addresses.


x/api/defaults_darwin.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !windows
+//go:build darwin
```
docker-agent, line 15: **[LOW] Platform coverage gap: non-darwin/non-linux Unix platforms have no `DaemonSocketPath()` implementation**

The original `defaults_unix.go` used `//go:build !windows`, which implicitly covered FreeBSD, OpenBSD, illumos, and any other non-Windows platform. This PR replaces it with two explicitly-tagged files (`darwin` and `linux`), leaving every other Unix-like OS without an implementation of `DaemonSocketPath()` — the package would fail to compile on those platforms.

In practice, docker/secrets-engine targets Linux and macOS (darwin), so this is theoretical rather than immediately harmful. However, the narrowing of platform coverage introduces a latent regression relative to the prior code with no documented rationale for the exclusion.

If intentional, consider adding a comment to `defaults_darwin.go` (or a new `defaults_unix_other.go` with `//go:build !linux && !windows && !darwin`) explaining the scope decision, so future contributors understand why FreeBSD etc. are not supported.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The abstract Unix domain socket refactor is well-implemented. The `@`-prefix convention is correct for Go's abstract socket namespace, UID namespacing is sound, and the plugin config test properly skips filesystem setup for abstract addresses.

One minor note logged as an inline comment.

Benehiko (approved):

PR #569, opened 2026-06-29
fix: return empty map for empty keychain on Linux ls

joe0BAB:
`docker pass ls` against an empty keychain printed "secret not found" and exited 1 on Linux, instead of printing nothing and exiting 0. Listing an empty collection is a valid empty result, not a miss.

The Linux Secret Service backend's GetAllMetadata treated a zero-result collection search as store.ErrCredentialNotFound. Drop that guard and return an empty (non-nil) map with a nil error, matching the macOS and Windows backends. ErrCredentialNotFound stays reserved for single-item lookups (Get/Delete).

Add a fake-driven regression test for the empty-collection case.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix is narrow and correct. `GetAllMetadata` previously treated a zero-result collection search the same as a missing single-item lookup (`store.ErrCredentialNotFound`), causing `docker pass ls` on an empty keychain to exit 1 with "secret not found" on Linux. The PR removes that guard so an empty search produces an empty (non-nil) map with a nil error, matching the macOS and Windows backends.

**Key correctness points verified:**
- `make(map[store.ID]store.Secret, 0)` returns a non-nil empty map — the nil-map distinction is handled correctly.
- The loop over `itemPaths` is a no-op when empty; no dead code paths.
- Deferred session/connection cleanup is unaffected by the early-return removal.
- `ErrCredentialNotFound` remains reserved for single-item lookups (`Get`/`Delete`) as intended.
- `TestKeychainGetAllMetadataEmpty` correctly exercises the path using the established `fakeService{}` fake, with appropriate `NoError`, `NotNil`, and `Empty` assertions.

No bugs were introduced by this change.

Benehiko (approved):

Issue #56, opened 2025-06-27
Detect if Keychain can be used


Benehiko:
On Linux the keychain might not exist, some APIs over DBUS might be disabled or the keychain daemon might not be available.

We should have a way to detect this, give an appropriate error when it is not supported so that we can provide the user 
with an appropriate fallback.

Related: https://github.com/docker/sbx-releases/issues/180

---

## Headless / no-collection environments (expanded scope)

A concrete failure mode: the secret-service daemon **is** running, but no usable collection exists. This is common on headless Debian/Ubuntu (e.g. SSH session, no graphical/PAM login), where `gnome-keyring` is installed but:

- the `login` collection was never created/unlocked (PAM only does this on a graphical/PAM login), **and**
- no `default` alias is set.

### Observed error

```
WARN: unknown authentication error type encountered: retrieve default account metadata: Object does not exist at path "/"
ERROR: auth login failed: saving credentials: Object does not exist at path "/"
```

### Root cause

In `getDefaultCollection` (`store/keychain/keychain_linux.go`):

1. `login` collection is not in the collection list → fall through to the alias lookup.
2. `ReadAlias("default")` is called. Per the secret-service spec, when no collection is assigned to an alias the service returns the **null object path `/`**.
3. The guard `!defaultKeychainObjectPath.IsValid()` does **not** catch this — godbus `ObjectPath.IsValid()` returns `true` for `/` (it is a syntactically valid path; see `dbus.go` `if s == "/" { return true }`).
4. `/` is therefore returned as the collection path, and every subsequent D-Bus call (`isCollectionUnlocked`, `CreateItem`, search) targets `/` → `Object does not exist at path "/"`.

### Scope to address under this issue

- [ ] Treat the null path `/` returned by `ReadAlias` as "no default collection" rather than a valid collection (don't rely on `IsValid()` alone).
- [ ] Surface a clear, typed error when no usable collection exists (distinct from "daemon unavailable"), so callers can branch to a fallback.
- [ ] Detect the broader unavailability cases already described above (no daemon, DBUS APIs disabled).
- [ ] Define the fallback behaviour for headless environments (e.g. encrypted filestore, or guided keyring init).

### Workaround (user-side)

Initialise a default collection so a valid path exists:

```bash
# ensure daemon is running with the secrets component (headless)
eval "$(dbus-launch)"
eval "$(gnome-keyring-daemon --start --components=secrets)"

# creating the first item auto-creates the 'default' collection
[...]

PR #571, opened 2026-07-01
feat(keychain)!: eager backend availability detection, context-aware New

Benehiko:
## Problem

On Linux the keychain store has no way to tell the caller, at construction time,
that the keychain backend is unavailable or unsupported (e.g. WSL / headless
hosts with no D-Bus session bus, or desktops with no `gnome-keyring`/`kwallet`
running). The failure surfaces lazily on the first operation, which is awkward
for callers who want to fall back to another store.

## ⚠️ Breaking change

`keychain.New` (and the `docker-pass` `PassStore` helper that wraps it) now take
a `context.Context` as their **first argument**. The module version will be
bumped accordingly. In-repo callers are updated in this PR; downstream callers
must add a `ctx` argument.

## Change

`keychain.New` now eagerly verifies the backend is reachable before returning.
Callers detect it with a single `errors.Is` branch:

```go
st, err := keychain.New(ctx, group, name, factory)
if errors.Is(err, keychain.ErrKeychainUnavailable) {
    // backend unreachable on this host — fall back to another store
}
```

`ctx` bounds the probe: on Linux it is a single D-Bus round-trip, so pass a
context with a deadline (`context.WithTimeout`) to cap it. `ctx` governs
construction only — `New` does not retain it for later store operations, and on
macOS/Windows the probe is a no-op so `ctx` is unused.

## Design

- **Probe:** `org.freedesktop.DBus.NameHasOwner("org.freedesktop.secrets")` on
  the **bus daemon object** (`conn.BusObject()`) with `FlagNoAutoStart` — **not**
  `Collections()`. The daemon answers from its own name registry and never
  forwards the query to the Secret Service, so it is **prompt-safe and
  side-effect-free**: it can never reach `PromptAndWait`, never activates the
  backend, and never touches a collection or item. It catches both target cases:
  no session bus (dial fails) and no keyring daemon (`hasOwner == false`).
- **One exported sentinel** `ErrKeychainUnavailable`, declared in the
  cross-platform file mirroring the existing `ErrNoDefaultCollection` precedent.
  Two unexported causes (`errSessionBusUnavailable`, `errNoSecretServiceOwner`)
  are wrapped underneath for future promotion without behavioral change.
- **Distinct from `ErrNoDefaultCollection`:** the probe does not assert a
  collection exists, so a reachable-but-uninitialized keyring still passes `New`
  and surfaces `ErrNoDefaultCollection` lazily on first use, exactly as before.
- **Per-platform hook** `ensureAvailable(ctx)`: Linux probes; macOS/Windows are
  no-ops (no unreachable-at-construction concept — Windows' `ErrNoLogonSession`
  stays lazy, out of scope). Runs through the existing `newService` seam and
  closes its connection immediately (preserves the fresh-connection-per-op
  contract). The probe uses the caller's `ctx` directly (no library-imposed
  timeout).

## Trade-off (documented)

A D-Bus-activatable backend that is installed but not yet running registers no
owner, so `NameHasOwner` returns false and `New` reports `ErrKeychainUnavailable`
even though the first real operation would have auto-started it. This is the
deliberate price of prompt-safety and eager failure, and matches the two target
cases. See `store/docs/keychain/decision-logs.md`.

## Testing

- New unit tests through the existing `newService`/`fakeService` seam:
  `TestNewProbeSucceeds`, `TestNewProbeNoOwner`, `TestNewProbeNoSessionBus`,
  `TestNewProbeTransportError`, `TestNewProbeShape`,
  `TestNewProbeContextCanceled` (verifies the caller's `ctx` flows into the probe
  and its cancellation surfaces under `ErrKeychainUnavailable`).
- `TestKeychainClosesEveryConnection` updated to account for the probe's
  balanced open/close.
- Verified: `go build` + `go vet` + `golangci-lint` (0 issues) on
  **darwin / linux / windows**, `gofmt` clean, and **all Linux keychain tests
  pass against a live `gnome-keyring`** (`ubuntu-24-gnome-keyring` container),
  including the real-keyring integration tests exercising the real
  `NameHasOwner` probe.

## Docs

`store/docs/keychain/decision-logs.md` (new dated entry), `design.md`
(availability subsection), and `store/keychain/README.md` (availability note).

## Related issues

Relates to #56 ("Detect if Keychain can be used"). This PR addresses the
backend-unavailability detection called for there:

- [x] Detect the unavailability cases (no daemon / DBUS APIs disabled / no
  session bus) and surface a typed error — `ErrKeychainUnavailable`.
- [x] Surface a clear, typed error distinct from "daemon unavailable" for the
  no-usable-collection case, and treat the null path `/` from `ReadAlias` as
  "no default collection" — already shipped as `ErrNoDefaultCollection` /
  `resolveDefaultCollection`; this PR keeps `ErrKeychainUnavailable` explicitly
  distinct from it.
- [ ] Define the headless fallback behaviour (encrypted filestore / guided
  keyring init) — **out of scope here**: this PR gives callers the typed signal
  to branch on; wiring an actual fallback store is a caller/app-level follow-up.

Not using a `Closes` keyword because that last item remains open.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The eager backend availability detection is well-implemented. The D-Bus probe in `ensureAvailable` correctly uses `org.freedesktop.DBus.NameHasOwner` on the bus daemon object (not the Secret Service itself), making it prompt-safe and side-effect-free. Connection lifecycle is handled correctly with `defer service.Close()` covering all code paths. Error wrapping chains properly support `errors.Is` for `ErrKeychainUnavailable`, `errSessionBusUnavailable`, and `context.Canceled`. The `SecretServiceInterface` constant (`"org.freedesktop.secrets"`) doubles as the well-known bus name so `NameHasOwner` receives the correct argument. Triple `%w` in `fmt.Errorf` is valid under the project's Go 1.25.11 minimum. macOS/Windows no-ops are correct. Test coverage is thorough, including the context-cancellation path (`TestNewProbeContextCanceled`).

No issues found in the changed code.

joe0BAB (dismissed):

joe0BAB (approved):

PR #580, opened 2026-07-14
feat(ci): make sign-release dispatchable for pre-publish signing

joe0BAB:
Replace the release:published trigger with workflow_dispatch (tag input) so the release pipeline can invoke signing mid-release, once the raw Linux binaries have been staged on a draft release at that tag. Signing the binaries before the release is published lets the cosign bundles be embedded into the downstream .deb/.rpm, while the recorded Rekor signer identity stays this public repository.


.github/workflows/sign-release.yml
```diff
@@ -1,17 +1,18 @@
...
-# trigger further workflows), this `release` event fires as expected.
+# Keyless-signs the Linux binaries staged on the release/draft at the given tag,
+# uploading the cosign bundles back onto it. Runs here so the Rekor-recorded
+# signer identity is this repository. Invoked via workflow_dispatch by the
+# release pipeline once the binaries have been staged.
 
 on:
-  release:
-    types: [published]
+  workflow_dispatch:
```
docker-agent, line 10: **[MEDIUM/SECURITY] `workflow_dispatch` with free-form `tag` input and no caller restriction: any dispatcher can sign arbitrary release assets**

Switching from `release: published` to `workflow_dispatch` meaningfully broadens who can pull this trigger. Under `release: published`, signing fired automatically only when a release was published — an action gated by release-management write access. With `workflow_dispatch`, any actor who holds repository write access (including CI bots, automation tokens, or contributors who can push branches but don't manage releases) can manually invoke the workflow with any `tag` string they choose.

Combined with `id-token: write` (OIDC/Fulcio keyless signing identity) and `contents: write` (upload bundles back to releases), an unauthorized dispatcher could:
- Obtain a Rekor-recorded Sigstore signature for binaries they did not produce, under this repository's identity.
- Upload those bundles to any existing release, including ones already published.

**Suggested mitigation**: Add an [`environment` protection rule](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) (e.g. `environment: release-signing`) to the `sign` job. Environment protection lets you require manual approval and/or restrict which actors/branches are allowed to deploy into it, without needing a separate allow-list in the workflow YAML:

```yaml
jobs:
  sign:
    name: Sign Linux binaries
    runs-on: ubuntu-latest
    environment: release-signing   # <-- add this
    timeout-minutes: 15
```

This ensures the OIDC token and `contents: write` permission are only issued after an authorized reviewer approves the dispatch.

docker-agent (commented):

.github/workflows/sign-release.yml
```diff
@@ -31,7 +36,7 @@ jobs:
       - name: Sign raw Linux binaries (keyless cosign)
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          TAG: ${{ github.event.release.tag_name }}
+          TAG: ${{ github.event.inputs.tag }}
```
docker-agent, line 39: **[HIGH] Unvalidated `workflow_dispatch` tag input used in shell — injection risk if actor guard is bypassed**

`TAG` is set directly from `${{ github.event.inputs.tag }}` (line 39) and then interpolated into two shell commands:

```bash
gh release download "$TAG" ...
gh release upload "$TAG" ...
```

The previous trigger (`release:published`) sourced the tag from a trusted GitHub event (`github.event.release.tag_name`). The new `workflow_dispatch` replaces that with free-form user input. Although `"$TAG"` prevents word-splitting, shell metacharacters such as `$(...)` or backticks embedded in the value can still be evaluated by the shell when the expanded env var is processed. The actor check and `environment: release-signing` gate reduce the risk, but neither is cryptographically enforced.

**Suggested fix** — add a format guard at the top of the `run:` block:
```bash
[[ "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9._-]+)?$ ]] \
  || { echo "::error::invalid tag format: $TAG"; exit 1; }
```
This is a cheap, explicit check that eliminates the injection surface entirely and makes the expected input contract visible in the code.

.github/workflows/sign-release.yml
```diff
@@ -21,6 +22,10 @@ jobs:
   sign:
     name: Sign Linux binaries
     runs-on: ubuntu-latest
+    # Restrict signing to the release automation's app: any repo-write actor can
+    # otherwise dispatch this and mint a keyless signature under this repo's identity.
+    if: github.triggering_actor == 'docker-read-write[bot]'
```
docker-agent, line 27: **[MEDIUM] `triggering_actor` string-match is the sole automated authorization gate; fragile**

```yaml
if: github.triggering_actor == 'docker-read-write[bot]'
```

Any GitHub user or app with `repo:write` (or `workflow`) permission can trigger a `workflow_dispatch` via the API. The `triggering_actor` check is the only automated guard preventing an unauthorized actor from initiating a cosign/Rekor signing run under this repository's OIDC identity. If the bot is renamed, the check silently stops working — the job is skipped with no workflow-level failure, so the release pipeline could silently proceed without signing without raising an alert.

The old `release:published` trigger was gated by the release event itself (not user-supplied). This new pattern introduces reliance on a mutable, human-readable string.

**Suggested hardening:**
1. Ensure `environment: release-signing` has required reviewers configured — make it the primary authorization gate, not just defense-in-depth.
2. Change the `if:` to fail loudly (e.g., add a separate explicit-failure step) so a dispatch by an unexpected actor is surfaced as a workflow error rather than a silent skip.
3. Document the exact expected bot actor name in a comment so regressions are caught in review.

docker-agent (commented):

The change is structurally sound — replacing `release:published` with `workflow_dispatch` enables pre-publish signing, which is the stated goal. Two concerns introduced by the PR warrant attention before merge:

1. **[HIGH]** The `tag` input is free-form user text used directly in shell commands without format validation (see inline comment at line 39).
2. **[MEDIUM]** The `triggering_actor` string check is the only automated authorization gate for minting Rekor signatures under this repo's OIDC identity; it's fragile and fails silently (see inline comment at line 27).

Neither issue is insurmountable — a one-line regex guard addresses (1), and tightening the `environment: release-signing` configuration addresses (2).

Benehiko (approved):

PR #581, opened 2026-07-14
feat(posixage): expose scrypt work factor as an option

Benehiko:
## Summary

Exposes the scrypt work factor (2^logN) used for password-protected secrets in the `posixage` store as a configurable option, `posixage.WithScryptWorkFactor(logN int)`.

Previously the work factor was fixed at age's `NewScryptRecipient` default (`logN=18`, ~1s per operation). Callers can now trade encryption/decryption time against brute-force resistance.

## Not a breaking change

The scrypt work factor is recorded in each file's `scrypt` stanza header (`-> scrypt <salt> <logN>`) and read back from that header on decryption. Changing the configured factor therefore:

- does **not** affect the ability to decrypt secrets written with a previous factor — each file carries its own;
- keeps new files decryptable by standard `age` tooling, provided `logN` stays at or below age's default decryption ceiling.

Values are validated to `1..22` (`secretfile.MaxScryptWorkFactor`); 22 matches age's default `maxWorkFactor`, so files never exceed what a stock age reader accepts. Out-of-range values are rejected at `New()` time rather than panicking (`age.ScryptRecipient.SetWorkFactor` panics outside `1..30`).

## Details

- `secretfile`: new variadic `RecipientOption` + `WithScryptWorkFactor`; `GetRecipients` applies it only to `PasswordKeyType` (no-op for age/ssh). Zero value leaves age's default in place, so existing callers are unaffected.
- `posixage`: new store-level `WithScryptWorkFactor` option, threaded into the encrypt path.
- Signatures stay backward-compatible (variadic).

## Testing

New `TestScryptWorkFactor` covers:

- configured factor is recorded in the header and round-trips;
- unset defaults to `logN=18`;
- a file written with a custom factor decrypts on a store with **no** work-factor option configured (proves the header is self-describing);
- out-of-range values (`0`, `-1`, `23`) are rejected by `New()`.

`gofmt` and `go vet` clean. `make lint`/`make format` not run locally (Docker registry keychain locked in this environment) — CI will run them.


store/posixage/store_test.go
```diff
@@ -799,3 +801,204 @@ func TestPOSIXAge(t *testing.T) {
...
+
+		// A separate store with no work-factor option configured.
+		reader := newPasswordStore(t, root, password)
+
+		got, err := reader.Get(t.Context(), id)
+		require.NoError(t, err)
+		assert.EqualValues(t, secret, got)
+	})
+
+	t.Run("rejects out-of-range work factors", func(t *testing.T) {
```
docker-agent, line 929: **[LOW] Boundary values logN=1 and logN=22 are not explicitly tested as valid accepted inputs**

The `rejects out-of-range work factors` sub-test covers `{0, -1, MaxScryptWorkFactor+1}` but does not verify that the boundary values themselves (logN=1 and logN=22) are accepted without error. An off-by-one regression — e.g. `logN < 1` becoming `logN <= 1`, or `logN > MaxScryptWorkFactor` becoming `logN >= MaxScryptWorkFactor` — would go undetected by the current test suite.

Consider adding a parallel acceptance loop:
```go
for _, logN := range []int{1, secretfile.MaxScryptWorkFactor} {
    _, err := New(root, ..., WithScryptWorkFactor(logN))
    assert.NoError(t, err, "logN=%d should be accepted", logN)
}
```

docker-agent (commented):
### Assessment: 🟢 APPROVE

The scrypt work-factor option is correctly threaded through the encrypt path, the zero-value bypass safely delegates to age's own default, validation fires at `New()` time (avoiding the age library panic), and the work factor is applied only to `PasswordKeyType` recipients. The self-describing header means changing the factor never breaks decryption of older files. No data-corruption, nil-pointer, or logic bugs were found in the changed code.

One minor test gap is noted inline: the boundary values `logN=1` and `logN=22` (MaxScryptWorkFactor) are not verified as accepted inputs, so a future off-by-one regression would go undetected.

joe0BAB (approved):

PR #583, opened 2026-07-14
ci: sign release binaries on the release publish event

joe0BAB:
Trigger keyless signing when a release is published (instead of a manual workflow_dispatch) and upload the cosign bundles back onto the release. Clarify the triggering behavior in comments.


.github/workflows/sign-release.yml
```diff
@@ -22,10 +21,6 @@ jobs:
   sign:
     name: Sign Linux binaries
     runs-on: ubuntu-latest
-    # Restrict signing to the release automation's app: any repo-write actor can
-    # otherwise dispatch this and mint a keyless signature under this repo's identity.
-    if: github.triggering_actor == 'docker-read-write[bot]'
```
docker-agent, line 27: **[HIGH] Actor guard removed — any collaborator publishing a release can trigger keyless signing under the repo's OIDC identity**

The `if: github.triggering_actor == 'docker-read-write[bot]'` guard was explicitly present to prevent any repo-write collaborator from dispatching this workflow and minting a keyless cosign signature under this repository's Sigstore/Fulcio identity. That guard is removed here without an in-workflow replacement.

With the new `release: types: [published]` trigger, the workflow fires for **every** published release, regardless of who published it. Any collaborator with `write` or higher access can publish a release through the GitHub UI or API — including one with arbitrary attached assets — which will:
1. Download whatever assets are on that release
2. Sign them with keyless cosign using this repo's OIDC identity
3. Upload the signature bundles back to the release

The workflow comment explains that the *intended* publisher is the GitHub App, but this is a convention, not an enforcement. A malicious or mistaken collaborator could get arbitrary binaries signed under this repository's trusted Rekor identity, potentially misleading downstream consumers about binary provenance.

**Suggested fix**: Add a job-level `if:` condition to enforce the restriction in-workflow, e.g.:
```yaml
if: github.triggering_actor == 'docker-read-write[bot]'
```
or, if the actor name differs for the `release` event context, use the `environment: release-signing` gate with required reviewers as an approval gate instead.

.github/workflows/sign-release.yml
```diff
@@ -22,10 +21,6 @@ jobs:
...
-    environment: release-signing
```
docker-agent, line 28: **[HIGH] `environment: release-signing` protection removed, eliminating any approval gate on the signing job**

The `environment: release-signing` declaration was removed. GitHub Actions environments can be configured with required reviewers, deployment branch restrictions, and wait timers — all of which serve as approval gates before a job runs.

Removing this environment declaration means the signing job now runs immediately and unconditionally when the `release: types: [published]` event fires, with no opportunity for a human to review or block the run. Combined with the removal of the actor guard (see the related finding), there is no remaining in-workflow safeguard.

If `release-signing` had required reviewers or branch restrictions configured, those protections are now bypassed entirely. Even if the environment currently has no rules, its removal forfeits the option to add them without a workflow change.

**Suggested fix**: Restore `environment: release-signing` on the signing job, and ensure the environment is configured with appropriate protection rules (e.g., required reviewers):
```yaml
jobs:
  sign:
    name: Sign Linux binaries
    runs-on: ubuntu-latest
    environment: release-signing
    timeout-minutes: 15
```

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR removes two security controls from the keyless cosign signing workflow without in-workflow replacements. Both findings are anchored on deleted lines that were explicitly present to prevent unauthorized signing.

Benehiko (approved):

PR #589, opened 2026-07-17
feat(plugins): add AccessControlModule plugin metadata type

joe0BAB:
Add an empty AccessControlModule message and wire it into the metadata oneof of RegisterPluginRequest and Plugin so plugins can register as access control modules.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes look mechanically correct. The proto field numbers match the generated Go code (, ), all  indices are correctly shifted to accommodate the new  at index 6,  and  are consistent,  are registered for both  and , and the case constants (, ) correspond to the correct proto field numbers. No bugs introduced.

Benehiko (approved):

PR #590, opened 2026-07-17
feat(accesscontrol): add Go access control interface and signing info types

joe0BAB:
Add the internal accesscontrol package that mirrors the accesscontrol.v1 wire API for use inside the engine:

- AccessControl interface with CheckAccess, taking a CheckAccessRequest (resolver pattern + process/signing context) and returning an allow/deny result plus error.
- ProcessInfo and a shared SigningInfoBase (SignedByDocker).
- Platform-specific SigningInfo built per GOOS:
  - darwin: SecCode* identity (team ID, identifier, CD hash) plus the dynamic CodeStatus bitmask.
  - windows: Authenticode signature identity, EV flag, mandatory IntegrityLevel, and untrusted PE version metadata.
  - linux: sigstore Fulcio identity/provenance and Rekor inclusion, one Signer entry per signature.


x/api/accesscontrol/accesscontrol.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+type AccessControl interface {
+	CheckAccess(ctx context.Context, req CheckAccessRequest) (bool, error)
+}
+
+type CheckAccessRequest struct {
+	secrets.Pattern
+	ProcessInfo
+	SigningInfo
```
docker-agent, line 16: **[MEDIUM] No fallback `SigningInfo` for non-darwin/linux/windows platforms — compile failure on other OSes**

`CheckAccessRequest` embeds `SigningInfo` (defined via filename-based build constraints in `_darwin.go`, `_linux.go`, and `_windows.go`), but `accesscontrol.go` carries no build constraint of its own. Any platform outside that set — `freebsd`, `openbsd`, `plan9`, `solaris`, etc. — will see an undefined `SigningInfo` identifier and fail to compile.

**Suggested fix**: Add a stub file `accesscontrol_unsupported.go` with a build tag covering the gap:

```go
//go:build !darwin && !linux && !windows

package accesscontrol

// SigningInfo is not supported on this platform.
type SigningInfo struct {
    SigningInfoBase
}
```

This keeps the package compilable on all platforms while making the "unsupported" case explicit.

docker-agent (commented):

This PR introduces a well-structured and thoroughly commented Go `accesscontrol` package. The code is clear, the platform-specific struct fields are well-documented with appropriate trust caveats, and the overall design is sound.

One confirmed build issue and one low-severity design ambiguity were found, both introduced by this PR.

x/api/accesscontrol/accesscontrol_windows.go
```diff
@@ -0,0 +1,92 @@
...
+
+// IntegrityLevel is a Windows mandatory integrity level, ordered so that higher
+// values denote a more privileged context. Mirrors winnt.h.
+type IntegrityLevel uint32
+
+const (
+	// IntegrityUntrusted is SECURITY_MANDATORY_UNTRUSTED_RID (0x0000). Also the
+	// zero value, used when the integrity level could not be determined.
+	IntegrityUntrusted IntegrityLevel = 0x0000
+	// IntegrityLow is SECURITY_MANDATORY_LOW_RID (0x1000), e.g. AppContainer or
```
docker-agent, line 80: **[LOW] `IntegrityUntrusted` (0x0000) doubles as "unknown" — ambiguous zero value**

The comment acknowledges this dual meaning ("Also the zero value, used when the integrity level could not be determined"), but the ambiguity is structural: a caller comparing `Integrity == IntegrityUntrusted` cannot distinguish a genuinely untrusted-labeled process from one where collection failed.

In practice, treating an unreadable integrity level as `IntegrityUntrusted` is a conservative/safe default (it denies rather than grants elevated access), but it could produce misleading diagnostic output and confuse future callers who are not aware of the documented dual meaning.

**Consider**: adding a separate `IntegrityUnknown` sentinel (e.g. `0xFFFF`) or a companion `IntegrityResolved bool` field to make the "we couldn't read it" case unambiguous.

Benehiko (approved):

PR #594, opened 2026-07-20
feat(accesscontrol): add connect handler and platform signing info mapping

joe0BAB:
Add accessControlService, which adapts an AccessControl implementation to
the generated connect AccessControlServiceHandler. CheckAccess parses the
requested secret pattern, builds a CheckAccessRequest from the requester's
process and signing info, and maps the boolean result to an ALLOW/DENY
Decision, returning InvalidArgument on bad patterns and Internal on check
failures.

Provide per-platform signingInfoFromProto implementations that translate
the protobuf Requester into the platform-specific SigningInfo:
- darwin: team ID, identifier, org, common name, CD hash, code status
- linux:  Sigstore signers (cert identity/issuer, source repo/ref/commit,
          runner environment, build trigger, Rekor log index, timestamp)
- windows: certificate chain trust, subject org/CN, issuer, SHA-256
          thumbprint, EV flag, integrity level, and file metadata


docker-agent (commented):

Two medium-severity issues found in the new access control service implementation. Both are in added code introduced by this PR.

x/api/accesscontrol/service_windows.go
```diff
@@ -0,0 +1,38 @@
...
+	if w == nil {
+		return si
+	}
+	si.TrustedChain = w.GetTrustedChain()
+	si.SubjectOrg = w.GetSubjectOrg()
+	si.SubjectCommonName = w.GetSubjectCommonName()
+	si.Issuer = w.GetIssuer()
+	si.ThumbprintSHA256 = w.GetThumbprintSha256()
+	si.IsEV = w.GetIsEv()
+	si.Integrity = IntegrityLevel(w.GetIntegrity())
```
docker-agent: **[MEDIUM] Direct int32→uint32 cast of proto `IntegrityLevel` enum silently produces bogus value for unknown/negative inputs**

`IntegrityLevel(w.GetIntegrity())` casts a proto `int32`-backed enum directly to the Go `IntegrityLevel` type (which is `uint32`-based). For the defined enum values (`0`, `0x1000`, `0x2000`, `0x3000`, `0x4000`) the cast is correct. However, if an unknown or future negative proto enum value arrives over the wire (e.g., `-1`), Go's numeric conversion produces `uint32(0xFFFFFFFF)` — not a valid `IntegrityLevel` constant. Downstream switch statements or comparisons on `Integrity` would silently mismatch, with no error returned and no validation performed.

Consider an explicit enum→constant mapping with a safe default:
```go
switch w.GetIntegrity() {
case accesscontrolv1.WindowsSigningInfo_INTEGRITY_LEVEL_LOW:
    si.Integrity = IntegrityLow
case accesscontrolv1.WindowsSigningInfo_INTEGRITY_LEVEL_MEDIUM:
    si.Integrity = IntegrityMedium
// ...
default:
    si.Integrity = IntegrityUntrusted // or return an error
}
```

x/api/accesscontrol/service.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (s accessControlService) CheckAccess(ctx context.Context, c *connect.Request[accesscontrolv1.CheckAccessRequest]) (*connect.Response[accesscontrolv1.CheckAccessResponse], error) {
+	msgPattern := c.Msg.GetPattern()
+	pattern, err := secrets.ParsePattern(msgPattern)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("invalid pattern %q: %w", msgPattern, err))
+	}
+
+	requester := c.Msg.GetRequester()
```
docker-agent, line 47: **[MEDIUM] Missing nil/absent check on `Requester` field — zero-value `ProcessInfo` and `SigningInfo` silently reach `CheckAccess`**

`c.Msg.GetRequester()` returns `nil` when the caller omits the `requester` field entirely. Proto getters on a nil receiver return zero values safely (no panic), so the resulting `CheckAccessRequest` carries `PID=0`, empty `Name`, empty `AbsoluteBinaryPath`, and a zero-value `SigningInfo` — effectively no process identity whatsoever. This bypasses the `InvalidArgument` guard (which only fires for a bad pattern) and lets a structurally invalid request reach `s.ac.CheckAccess`. If the `CheckAccess` backend can return `allowed=true` for a zero-identity request, this becomes a security issue.

Consider adding an explicit guard:
```go
requester := c.Msg.GetRequester()
if requester == nil {
    return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("requester is required"))
}
```

x/api/accesscontrol/service_windows.go
```diff
@@ -0,0 +1,53 @@
...
+	if w == nil {
+		return si
+	}
+	si.TrustedChain = w.GetTrustedChain()
+	si.SubjectOrg = w.GetSubjectOrg()
+	si.SubjectCommonName = w.GetSubjectCommonName()
+	si.Issuer = w.GetIssuer()
+	si.ThumbprintSHA256 = w.GetThumbprintSha256()
+	si.IsEV = w.GetIsEv()
+	si.Integrity = integrityLevelFromProto(w.GetIntegrity())
```
docker-agent, line 33: **[MEDIUM] Unvalidated cast of proto IntegrityLevel enum may pass unknown values into trust logic**

`w.GetIntegrity()` returns a `WindowsSigningInfo_IntegrityLevel` proto3 enum, which permits any integer value on the wire — a client can transmit a value that isn't defined in the proto enum. The direct cast `IntegrityLevel(w.GetIntegrity())` forwards that unknown integer unchanged into `si.Integrity` and from there into the access-control policy engine.

If the downstream `CheckAccess` implementation evaluates `IntegrityLevel` via an exhaustive switch or range check, an unrecognised value could fall into a permissive default case, elevating an untrusted process above its actual Windows integrity level. Consider adding an explicit validation step before the cast, e.g.:

```go
switch w.GetIntegrity() {
case accesscontrolv1.WindowsSigningInfo_INTEGRITY_LEVEL_LOW,
     accesscontrolv1.WindowsSigningInfo_INTEGRITY_LEVEL_MEDIUM,
     accesscontrolv1.WindowsSigningInfo_INTEGRITY_LEVEL_HIGH,
     accesscontrolv1.WindowsSigningInfo_INTEGRITY_LEVEL_SYSTEM:
    si.Integrity = IntegrityLevel(w.GetIntegrity())
default:
    return si, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("unknown integrity level: %d", w.GetIntegrity()))
}
```

x/api/accesscontrol/service_darwin.go
```diff
@@ -0,0 +1,34 @@
...
+	d := r.GetDarwin()
+	if d == nil {
+		return si
+	}
+	si.TeamID = d.GetTeamId()
+	si.Identifier = d.GetIdentifier()
+	si.Organization = d.GetOrganization()
+	si.CommonName = d.GetCommonName()
+	si.CDHash = d.GetCdHash()
+	si.Status = CodeStatus(d.GetStatus())
```
docker-agent, line 32: **[MEDIUM] Unvalidated bitmask cast for CodeStatus forwards arbitrary bits into trust-sensitive Darwin signing logic**

`d.GetStatus()` returns a raw `uint32` from the proto (the field is typed as `uint32`, not an enum). The direct cast `CodeStatus(d.GetStatus())` accepts any bit pattern a client sends, including bits that are not defined as constants in `CodeStatus`. These undefined bits are then forwarded verbatim into `si.Status` and from there into the access-control policy engine.

If the downstream `CheckAccess` implementation checks individual status flags (e.g., `StatusDebugged` as a denial signal), a client that sets both a defined flag and undefined bits could confuse bitmask logic. The safe fix is to mask to the known bits before assigning:

```go
const knownStatusBits = StatusValid | StatusHard | StatusKill | StatusDebugged
si.Status = CodeStatus(d.GetStatus()) & knownStatusBits
```

This ensures no out-of-spec bits survive the translation layer.

docker-agent (commented):

Two medium-severity findings in the platform-specific signing-info translators. The core service logic (service.go) and the Linux/Windows/Darwin field mappings are otherwise correct and complete.

**Findings summary:**
- `service_windows.go:33` — proto3 enum cast to IntegrityLevel without validation; unknown enum values pass unchecked into trust logic
- `service_darwin.go:32` — unvalidated uint32 bitmask cast to CodeStatus; arbitrary bits from a malformed request reach the access-control policy engine

One high-severity hypothesis (nil-dereference on absent IntegratedTime) was investigated and dismissed: `timestamppb.Timestamp.AsTime()` is safe to call on a nil receiver because the underlying `GetSeconds()`/`GetNanos()` getters are nil-safe; an absent field yields the Unix epoch rather than a panic.

Benehiko (commented):
Automated inline review of the connect handler + platform signing-info mapping (HEAD f20471e). Findings inline, most severe first: one correctness bug (nil timestamp → epoch), one trust-model question to confirm, plus robustness / coverage / style notes. Nice defensive touch replacing the raw enum cast with `integrityLevelFromProto` — that mapping is correct and fail-safe to `IntegrityUntrusted`.

x/api/accesscontrol/service_linux.go
```diff
@@ -0,0 +1,42 @@
...
+			CertIdentity:      s.GetCertIdentity(),
+			SourceRepo:        s.GetSourceRepo(),
+			SourceRef:         s.GetSourceRef(),
+			SourceCommit:      s.GetSourceCommit(),
+			RunnerEnvironment: s.GetRunnerEnvironment(),
+			BuildTrigger:      s.GetBuildTrigger(),
+			RunInvocationURI:  s.GetRunInvocationUri(),
+			RekorLogIndex:     s.GetRekorLogIndex(),
+			IntegratedTime:    s.GetIntegratedTime().AsTime(),
+		})
```
Benehiko: **Correctness — nil `integrated_time` maps to epoch, not the zero `time.Time`.** `GetIntegratedTime()` returns `nil` when the field is unset, and `(*timestamppb.Timestamp)(nil).AsTime()` returns `time.Unix(0, 0).UTC()` (1970-01-01) rather than the zero `time.Time`. A `Signer` with no Rekor time then has `IntegratedTime.IsZero() == false`, so any staleness/audit check treats an absent time as a real 1970 timestamp. Suggest guarding: only call `.AsTime()` when `s.GetIntegratedTime() != nil`, otherwise leave the zero value.

x/api/accesscontrol/service.go
```diff
@@ -0,0 +1,73 @@
...
```
Benehiko, line 47: **Trust boundary — signing fields are trusted verbatim from the wire.** `signed_by_docker` and every platform signing field are copied straight from the request into the access decision with no re-verification here. If this endpoint is reachable by the requesting peer itself, a malicious process can send `signed_by_docker: true` (or a forged `team_id`) and any `SignedByDocker`-gated policy will grant it. The deleted proto comment ("the verified context of the peer process") implies a trusted engine-side collector must populate `Requester` before this RPC. Please confirm that guarantee is enforced upstream — if it isn't, this is a critical trust bypass.

x/api/accesscontrol/service_linux.go
```diff
@@ -0,0 +1,42 @@
...
+package accesscontrol
+
+import accesscontrolv1 "github.com/docker/secrets-engine/x/api/accesscontrol/v1"
+
+func signingInfoFromProto(r *accesscontrolv1.Requester) SigningInfo {
+	si := SigningInfo{
+		SigningInfoBase: SigningInfoBase{SignedByDocker: r.GetSignedByDocker()},
+	}
+	l := r.GetLinux()
+	if l == nil {
```
Benehiko, line 24: **Robustness — an unset or mismatched `signing_info` oneof silently drops all trust fields.** When the oneof is unset (or holds a different platform's variant), `GetLinux()` returns `nil` and only `SignedByDocker` survives; `Signers` and the rest become zero values with no error or log. The access decision is then made on empty signing context — a policy requiring a specific signer silently denies a legitimate peer, and a `SignedByDocker`-only policy never learns the identity was missing. Consider returning `InvalidArgument` (or at least logging) when a variant is expected but absent. Same pattern applies in `service_darwin.go` and `service_windows.go`.

x/api/accesscontrol/service.go
```diff
@@ -0,0 +1,73 @@
...
+
+var _ accesscontrolv1connect.AccessControlServiceHandler = &accessControlService{}
+
+type accessControlService struct {
+	ac AccessControl
+}
+
+// NewAccessControlHandler adapts an AccessControl implementation to the
+// generated connect handler interface.
+func NewAccessControlHandler(ac AccessControl) accesscontrolv1connect.AccessControlServiceHandler {
```
Benehiko, line 36: **Test coverage — the package ships no tests.** Untested surface: the four `CheckAccess` branches (invalid pattern, nil requester, `ac` error, allow/deny mapping) and the three platform `signingInfoFromProto` mappers (plus the new `integrityLevelFromProto`). A regression that inverts the allow/deny mapping or drops a field in a mapper would ship undetected. Worth adding table tests — at minimum asserting `allowed == true → DECISION_ALLOW` and `nil requester → CodeInvalidArgument`.

x/api/accesscontrol/service.go
```diff
@@ -0,0 +1,73 @@
...
+	return &accessControlService{ac}
+}
+
+func (s accessControlService) CheckAccess(ctx context.Context, c *connect.Request[accesscontrolv1.CheckAccessRequest]) (*connect.Response[accesscontrolv1.CheckAccessResponse], error) {
```
Benehiko: **Minor — value receiver on a pointer-only type.** `CheckAccess` uses a value receiver while the type is only ever used by pointer (`NewAccessControlHandler` returns `&accessControlService{}`, and the compile-time assertion uses `&accessControlService{}`). Each call copies the struct, and it diverges from the constructor. Prefer a pointer receiver for consistency.

x/api/accesscontrol/service.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2025-2026 Docker, Inc.
```
Benehiko, line 1: **Nit — copyright year inconsistent with the rest of the package.** These four new files use `Copyright 2025-2026 Docker, Inc.`, while every existing file in the package (including the headers added in 4649fea) uses `Copyright 2026 Docker, Inc.`. Applies to `service.go`, `service_darwin.go`, `service_linux.go`, and `service_windows.go`.

Benehiko (approved):

PR #595, opened 2026-07-20
feat(accesscontrol): wire AccessControlModule into plugin setup

joe0BAB:
Complement the SecretsProvider switch case in plugin setup with an AccessControlModule branch. Add the connect handler adapter and platform-specific proto->domain signing-info mapping, plus a registration-gating wrapper mirroring the resolver service.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly wires `AccessControlModule` into the plugin setup alongside the existing `SecretsProvider` path. The type switch in `setup.go`, the new `NewAccessControlModule` constructor, the registration proto builder, and the `Config.Valid()` validator all follow the existing patterns consistently.

All hypotheses raised during review were dismissed on verification:
- The silent no-op in the type switch and the `any` parameter in `newStub` are protected by Go's type system — the public constructors (`NewSecretsProvider`/`NewAccessControlModule`) accept typed interfaces, so only correct values reach `setup()`.
- The `AccessControlModule_builder{}.Build()` with no fields is consistent with the AccessControlModule proto definition having no required identifying fields analogous to SecretsProvider's pattern.

**Low-severity observations** (no action required):
- `time.After` inside `registrationGate.await` (inherited from the existing resolver pattern) leaks timer goroutines until they fire when concurrent requests arrive before setup completes — this is a known Go idiom trade-off and mirrors the pre-existing resolver service behavior.
- The timeout test case in `lifecycle_test.go` uses `registrationTimeout: 0` which fires immediately; this is technically nondeterministic but in practice works due to the timer firing atomically before the goroutine can close `setupCompleted`.

plugin/config.go
```diff
@@ -69,13 +69,13 @@ func WithConnection(conn net.Conn) ManualLaunchOption {
 
 type cfg struct {
 	Config
-	plugin              ExternalPlugin
+	plugin              any
```
Benehiko: i don't like `any`. 
Is there no way to have a generic or interface here?

plugin/setup.go
```diff
@@ -39,11 +44,16 @@ func setup(ctx context.Context, config cfg, onClose func(err error)) (io.Closer,
...
 	}}))
 	setupCompleted := make(chan struct{})
-	httpMux.Handle(resolverv1connect.NewResolverServiceHandler(&resolverService{
-		handler:             resolverv1.NewResolverHandler(config.plugin),
-		setupCompleted:      setupCompleted,
-		registrationTimeout: config.registrationTimeout,
-	}))
+	gated := connect.WithInterceptors(setupInterceptor(setupCompleted, config.registrationTimeout))
+	switch v := config.plugin.(type) {
+	case SecretsProvider:
```
Benehiko: **Correctness — handler dispatch and registration use different signals.**

`setup()` selects the service handler by an interface type-switch (`SecretsProvider` tested first), while `register.go` declares the module from the config (`SecretsProviderConfig` / `AccessControlConfig`). These two decisions can disagree.

If a plugin passed to `NewAccessControlModule` has a concrete type that *also* implements `GetSecrets` (both `CheckAccess` and `GetSecrets`), then:
- `register.go` sends `AccessControlModule` (config-driven), so the engine registers it as an access-control module and calls `CheckAccess`.
- `setup.go` matches `case SecretsProvider` first and mounts **only** the resolver handler.

Result: the engine calls `CheckAccess` but the plugin never serves the access-control service → unimplemented / 404.

Suggest driving this switch off the same config signal `register.go` uses (`config.SecretsProviderConfig != nil` / `config.AccessControlConfig != nil`) so registration and served handler can never diverge.

plugin/stub.go
```diff
@@ -48,21 +48,25 @@ var (
...
-	return nil
+	return newStub(p, config, opts...)
 }
 
-// New creates a stub with the given plugin and options.
-// ManualLaunchOption only apply when the plugin is launched manually.
-// If launched by the secrets runtime, they are ignored.
-// If logger is nil, a default logger will be created and used.
-func New(p ExternalPlugin, config Config, opts ...ManualLaunchOption) (Stub, error) {
+func newStub(p any, config Config, opts ...ManualLaunchOption) (Stub, error) {
```
Benehiko: **Correctness (pre-existing, in this touched function) — stub timeout accessors always return 0.**

`newStub` builds `&stub{name, factory}` (around line 81) and never sets `registrationTimeout` / `requestTimeout`, so `Stub.RegistrationTimeout()` and `Stub.RequestTimeout()` always return `0` — not the default the `Stub` interface doc comment promises ("the default timeout if the plugin has not been started"). `cfg.registrationTimeout` is populated (default or `WithRegistrationTimeout`) but never copied into the stub, and the `RequestTimeout` returned in the `RegisterPlugin` response is never consumed either.

Fix: `stub{registrationTimeout: cfg.registrationTimeout, ...}` and plumb the response `RequestTimeout` back into the stub. Not introduced by this PR, but flagged since the function was renamed/edited here.

plugin/lifecycle.go
```diff
@@ -33,3 +36,18 @@ func (s *pluginService) Shutdown(ctx context.Context, _ *connect.Request[plugins
...
+
+func setupInterceptor(chSetup chan struct{}, timeout time.Duration) connect.UnaryInterceptorFunc {
+	return func(next connect.UnaryFunc) connect.UnaryFunc {
+		return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+			select {
+			case <-chSetup:
+				return next(ctx, req)
+			case <-ctx.Done():
+				return nil, connect.NewError(connect.CodeInternal, errors.New("context cancelled while waiting for registration"))
+			case <-time.After(timeout):
```
Benehiko, line 55: **Efficiency — a timer is allocated on every request, even after setup completes.**

`select` evaluates all channel operands before blocking, so `time.After(timeout)` allocates a `Timer` on *every* RPC. Once `chSetup` is closed the `<-chSetup` case wins immediately, but the timer still lives for the full `timeout` (e.g. the registration timeout, ~30s) before it can be GC'd. Under a high request rate this piles up pending timers.

A fast path avoids it once setup is done:
```go
select {
case <-chSetup:
    return next(ctx, req)
default:
}
// only reached before setup completes
select {
case <-chSetup:
    return next(ctx, req)
case <-ctx.Done():
    ...
case <-time.After(timeout):
    ...
}
```

plugin/lifecycle_test.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"connectrpc.com/connect"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSetupInterceptor(t *testing.T) {
```
Benehiko, line 26: **Test coverage — the gate is no longer tested against the real handlers.**

`TestSetupInterceptor` exercises the interceptor in isolation with a `next` that returns `(nil, nil)`. Nothing now verifies that `setup()` actually wires the interceptor onto the resolver / access-control handlers, or that a real `GetSecrets` passes through after setup — coverage the deleted `resolver_test.go` used to provide (`Test_setup` only checks shutdown).

A future change that dropped the `gated` option from `httpMux.Handle` in `setup.go`, or mis-wired the access-control handler, would pass all current tests. Consider a setup-level test that asserts requests are blocked before `setupCompleted` closes and flow through afterwards.

plugin/stub.go
```diff
@@ -48,21 +48,25 @@ var (
...
+func NewSecretsProvider(p SecretsProvider, config Config, opts ...ManualLaunchOption) (Stub, error) {
+	if config.SecretsProviderConfig == nil {
+		return nil, errors.New("secrets provider config is required")
 	}
-	if c.Pattern == nil {
-		return errors.New("pattern is required")
+	return newStub(p, config, opts...)
+}
+
+func NewAccessControlModule(p AccessControlModule, config Config, opts ...ManualLaunchOption) (Stub, error) {
```
Benehiko, line 62: **Docs — new public API is undocumented.**

`NewAccessControlModule` (and the new `AccessControlModule` / `AccessControlConfig` / `AccessControl` exported symbols) are user-facing API but have no doc comment, and the README/docs have no access-control example — unlike the sibling `NewSecretsProvider`, which has both. The README section was updated only for the secrets-provider flow.

Add a doc comment on `NewAccessControlModule` and a short README example for building an access-control plugin.

Benehiko (commented):
Automated code review (`/code-review`, xhigh effort). 5 findings, most-severe first — one correctness issue worth resolving before merge (#1), plus a pre-existing correctness bug now sitting in a touched function (#2), an efficiency nit, a test-coverage gap, and a docs gap. Details inline.

plugin/config.go
joe0BAB: see b32b262a06b1ad2ba8c8dab7ae153c730126c117

Benehiko (approved):

PR #598, opened 2026-07-22
feat(client): surface access control module metadata in PluginInfo

joe0BAB:
Add AccessControlModuleMetadata to PluginInfo and populate it from the Plugin.access_control_module oneof case in ListPlugins. Handle the metadata oneof with a switch on WhichMetadata() so the mutually exclusive cases (secrets provider vs. access control module) are explicit.


Benehiko (approved):

client/client.go
```diff
@@ -292,16 +295,19 @@ func (c client) DisablePlugin(ctx context.Context, name string) error {
...
+	Disabled            bool
+	External            bool
+	Configurable        bool
+	RunStatus           pluginsv1.RunStatus
+	StatusMessage       string
+	SecretsProvider     *SecretsProviderMetadata
+	AccessControlModule *AccessControlModuleMetadata
 }
 
+type AccessControlModuleMetadata struct{}
```
docker-agent, line 309: **[LOW] `AccessControlModuleMetadata` is an empty struct — protobuf fields from the `AccessControlModule` message are silently dropped**

`AccessControlModuleMetadata` is defined as `struct{}` and populated with `&AccessControlModuleMetadata{}` without reading any fields from the incoming `pluginsv1.AccessControlModule` protobuf message. If the proto message currently has, or later acquires, meaningful fields (e.g. policy version, realm bindings), those values will be silently discarded by this mapping layer.

If the intent is purely a presence-indicator ("this plugin *is* an ACM"), this is fine and worth a brief comment to that effect (e.g. `// AccessControlModuleMetadata is a marker type; the ACM protocol carries no additional per-plugin metadata.`). If the proto message does carry fields, consider mirroring them here the same way `SecretsProviderMetadata` mirrors `Pattern`.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The switch on `WhichMetadata()` correctly replaces the previous nil-check and is idiomatic for buf's opaque protobuf API. The `SecretsProvider` and `AccessControlModule` oneof cases are handled explicitly with no behavioral regression. One low-severity observation about the empty metadata struct is left as an inline comment.

PR #600, opened 2026-08-03
feat(accesscontrol): model macOS signing identity as root/leaf chain

joe0BAB:
Replace the flat Darwin SigningInfo with a chain-aware model:

- Anchor classifies the certificate chain's root (none, ad-hoc, other, apple-generic, apple-platform), ordered by increasing trust so consumers can gate on a minimum.
- SigningIdentity carries the per-process code-signing fields (team ID, identifier, organization, common name, cdhash, status) plus Anchor and its own SignedByDocker.
- SigningInfo now describes the requesting process's ancestry: Root and Leaf signing identities plus the full process Chain.
- ProcessNode records one ancestor: PID, start token (survives PID reuse), effective UID, comm, exe/real exe paths, mtime, and argv.

Mirror the same shape in the accesscontrol.v1 proto (DarwinSigningInfo with nested SigningIdentity, ProcessNode, and Anchor enum) and update the proto-to-Go conversion accordingly.


Benehiko (approved):

x/api/accesscontrol/service_darwin.go
```diff
@@ -17,18 +17,62 @@ package accesscontrol
...
-	si.Status = CodeStatus(d.GetStatus())
 	return si
 }
+
+func signingIdentityFromProto(p *accesscontrolv1.DarwinSigningInfo_SigningIdentity) *SigningIdentity {
+	if p == nil {
+		return nil
+	}
+	return &SigningIdentity{
+		SigningInfoBase: SigningInfoBase{SignedByDocker: p.GetSignedByDocker()},
```
docker-agent, line 54: **[MEDIUM] `SignedByDocker` is now read from `DarwinSigningInfo.SigningIdentity`, silently dropping `Requester.signed_by_docker`**

The old code read `SignedByDocker` from the top-level `Requester` message:
```go
// old
SigningInfoBase{SignedByDocker: r.GetSignedByDocker()}
```
The new code reads it from the nested `DarwinSigningInfo_SigningIdentity`:
```go
// new
SigningInfoBase{SignedByDocker: p.GetSignedByDocker()}
```
These are completely separate proto fields. Any sender that populates `Requester.signed_by_docker` (the old location) will have that value silently ignored by the new receiver — `SignedByDocker` will always be `false` for those callers, which could cause a trust decision to go the wrong way.

If the sender side is being updated atomically alongside this receiver change (or if no deployed sender sets `Requester.signed_by_docker` yet), this is safe. Otherwise, consider temporarily reading from both fields and preferring the new one, or document in a migration note which proto version bump makes the old field safe to drop.

docker-agent (commented):

This PR refactors the macOS signing info model from a flat structure to a chain-aware root/leaf design. The structural changes are well-executed: proto field numbers are consistent between `.proto` and the generated `pb.go`, the `Anchor` enum values are correctly ordered by increasing trust level, the Go type definitions mirror the proto shape faithfully, and the `Mtime` nil-guard in `signingInfoFromProto` correctly avoids the Unix-epoch trap from calling `AsTime()` on a nil `Timestamp`.

One finding warrants attention: `SignedByDocker` has moved from `Requester.signed_by_docker` (top-level) to `DarwinSigningInfo.SigningIdentity.signed_by_docker` (nested). The receiver now reads the nested field, so any sender that still sets the top-level field will have its value silently dropped — trust decisions that depend on `SignedByDocker` would incorrectly evaluate to `false`.

PR #605, opened 2026-08-03
fix(keychain)!: return empty map from Filter on Linux when nothing matches

### LINKED ISSUE #579: keychain: Linux Filter returns ErrCredentialNotFound on empty/no-match while macOS & Windows return an empty map
### Summary

The Linux Secret Service backend's `Filter` is the only platform that treats "no matching secrets" as an error (`store.ErrCredentialNotFound`). macOS and Windows return an empty (non-nil) map with a `nil` error for the same situation. This per-OS divergence in a shared interface method appears unnecessary and forces callers to special-case Linux.

### Where

`store/keychain/keychain_linux.go`, `Filter` — empty collection:

```go
if len(itemPaths) == 0 {
    return nil, store.ErrCredentialNotFound
}
```

...and again at the end of the same function when nothing matched the pattern:

```go
if len(credentials) == 0 {
    return nil, store.ErrCredentialNotFound
}
```

Compare to `keychain_darwin.go` `Filter`, which returns `make(map[store.ID]store.Secret)` + `nil` even when empty (no empty-as-error guard). Windows behaves the same as macOS.

The Linux backend itself already acknowledges this asymmetry is undesirable — `keychain_linux.go` `GetAllMetadata` deliberately returns an empty map on an empty collection, with the comment:

> An empty collection is a valid empty result for "list all", not a miss ... This matches the macOS and Windows backends, which have no empty-as-error guard.

`Filter` was apparently not given the same treatment.

### Why it's a problem

1. **Semantics mismatch.** `Filter` is a list/query operation. An empty result set is a normal, successful outcome — not a not-found condition. `Get` returning `ErrCredentialNotFound` makes sense; `Filter` doing so does not.
2. **Inconsistent interface contract.** Callers coding against `store.Store` cannot rely on uniform behavior across backends, which defeats the purpose of the abstraction and makes bugs invisible to developers on macOS/Windows.

### Expected behavior

`Filter` should return an empty, non-nil `map[store.ID]store.Secret` with a `nil` error when no secrets match, on **all** platforms — matching the existing behavior of `GetAllMetadata` and of the macOS/Windows `Filter` implementations.

### Proposed fix

Remove both `len(...) == 0 → ErrCredentialNotFound` guards from `keychain_linux.go` `Filter` and return the (empty) `credentials` map instead. Update/relax any Linux `Filter` tests that assert `ErrCredentialNotFound` on empty.

### Notes / risk

- This is a behavior change for any Linux caller that currently relies on `Filter` returning `ErrCredentialNotFound` on empty. Such callers should already tolerate an empty map (that's what they get on macOS/Windows), but existing callers should be grepped for `errors.Is(err, ErrCredentialNotFound)` around `Filter` before shipping.
- Consider documenting the empty-result contract explicitly on the `Store.Filter` interface method to prevent the three implementations from drifting again.

Benehiko:
## What

The Linux Secret Service backend's `Filter` returned `store.ErrCredentialNotFound` when nothing matched — both when the collection search returned no items and when items existed but none survived pattern matching. The macOS and Windows backends return an empty map with a nil error in the same situation.

This removes both empty-as-error guards from the Linux `Filter`, following the same rationale as 4a163cd, which aligned `GetAllMetadata` in the same way.

## Why

The `store.Store.Filter` godoc never specified what an empty result means, which is what allowed the platforms to diverge in the first place. This PR makes the contract explicit: a pattern matching nothing is a valid empty result, not a miss. `ErrCredentialNotFound` stays reserved for single-item lookups (`Get`/`Delete`).

Note: the only current caller (`plugins/pass` `GetSecrets`) maps an empty result to `plugin.ErrNotFound` itself, so end-to-end behavior through that path is unchanged. The fix matters for cross-platform consistency and future callers of `Filter`.

## Testing

- Fake-driven regression tests: empty collection, and items present but no pattern match (asserting the secret is never loaded).
- Full `make keychain-linux-unit-tests` bake matrix passes (Fedora 43 + Ubuntu 24, gnome-keyring).
- `golangci-lint fmt` and `run` clean.

Fixes #579


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR removes two early-return guards in `keychain_linux.go` that incorrectly returned `store.ErrCredentialNotFound` when `Filter` found no matching items — aligning the Linux Secret Service backend with the empty-result contract already followed by the macOS and Windows backends. A clarifying godoc comment is added to `store/store.go` to make this contract explicit.

All added lines are documentation comments or test assertions. No new logic paths are introduced; only error-returning paths are removed. No resource leaks, nil dereferences, or logic errors were found in the changed code.

joe0BAB (approved):

PR #606, opened 2026-08-03
fix(deps): bump grpc to v1.83.0 and otel to v1.44.0 for CVE fixes

Benehiko:
## What

Bumps `google.golang.org/grpc` to v1.83.0 and the `go.opentelemetry.io/otel` family to v1.44.0 across all workspace modules, with `go mod tidy` and `go work vendor`.

## Why

govulncheck flagged three reachable vulnerabilities (call paths in `x/ipc`, `x/telemetry`, `x/testhelper`):

- [GO-2026-6061](https://pkg.go.dev/vuln/GO-2026-6061) — grpc xDS RBAC authorization engine + HTTP/2 transport server (fixed in v1.82.1)
- [GO-2026-5506](https://pkg.go.dev/vuln/GO-2026-5506) — otel multi-value baggage header excessive allocations (fixed in v1.41.0)
- [GO-2026-5426](https://pkg.go.dev/vuln/GO-2026-5426) — otel/sdk BSD kenv PATH hijacking (fixed in v1.43.0)

## Verification

- `govulncheck ./...` re-run on all modules: 0 reachable vulnerabilities
- `go build ./...` passes in all six modules
- `go test ./...` passes in `x`
- Go toolchain already at latest 1.25.x patch (1.25.12) — no stdlib findings


vendor/google.golang.org/grpc/internal/transport/readyreader/ready_reader.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil
+	}
+	rr := &nonBlockingReader{raw: raw}
+	rr.doRead = func(fd uintptr) bool {
+		s := &rr.state
+
+		s.buf = s.pool.Get(s.bufSize)
+		s.bytesRead, s.readError = sysRead(fd, *s.buf)
+
+		if s.readError != nil {
```
docker-agent, line 100: **[MEDIUM] `nonBlockingReader.doRead` discards data when `sysRead` returns `(n>0, err!=nil)` simultaneously**

The `doRead` callback frees the buffer whenever any error is returned, even if bytes were also read:

```go
s.bytesRead, s.readError = sysRead(fd, *s.buf)

if s.readError != nil {
    s.pool.Put(s.buf)  // buffer freed — n bytes lost
    s.buf = nil
}
return !wouldBlock(s.readError)
```

When `sysRead` returns `(n > 0, io.EOF)` — a valid POSIX scenario on the last read of a gracefully-closing connection — the buffer is freed and `ReadOnReady` returns `nil, 0, io.EOF`, discarding the `n` bytes that were successfully received. Yet the `Reader` interface doc at the top of this file explicitly states: *"Callers should always process the n > 0 bytes returned before considering the error."* The `nonBlockingReader` implementation does not honour its own interface contract.

The `bufReadyReader.Read` doc also notes that a non-zero count with `io.EOF` is possible, but the non-blocking path silently drops that data. The `blockingReader` fallback (line 152) has the same issue, but it at least matches the standard `io.Reader` convention. Only the `nonBlockingReader` (used on TCP connections) drops bytes.

**Suggested fix**: when `s.readError != nil && s.bytesRead > 0`, keep the buffer and let `ReadOnReady` return `(buf, n, readErr)` so callers can process the data before handling the error, in line with the interface contract.

docker-agent (commented):

This PR correctly bumps grpc to v1.83.0 and otel to v1.44.0 to fix three CVEs. The version bumps themselves are consistent across all workspace modules (`client`, `plugin`, `plugins/credentialhelper`, `plugins/pass`, `x`). The CVE fix targets (grpc ≥ v1.82.1, otel ≥ v1.41.0/v1.43.0) are satisfied.

Two issues were found in the generated `vendor/modules.txt` and one in the new upstream `readyreader` package introduced by grpc v1.83.0:

**vendor/modules.txt — duplicate module headers (not shown in GitHub diff but present in the file):**
`vendor/modules.txt` contains duplicate `#` header lines for the same module paths:
- `github.com/docker/secrets-engine/plugin` appears twice (at `v0.0.0-00010101000000-000000000000 => ./plugin` and `v0.0.8 => ./plugin`)
- `github.com/docker/secrets-engine/x` appears twice (at `v0.0.0-00010101000000-000000000000 => ./x` and `v0.3.0-do.not.use => ./x`)
- A bare `github.com/docker/secrets-engine v0.0.7` entry appears with `## explicit` but no package paths

The Go toolchain expects each module path to appear exactly once in `vendor/modules.txt`. These duplicates suggest that `go mod tidy` was run independently in some workspace members with inconsistent version references before `go work vendor` was run, producing a malformed output. Run `go work vendor` from the workspace root after ensuring all member `go.mod` files agree on consistent versions for these local dependencies.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR correctly bumps `google.golang.org/grpc` to v1.83.0 and the `go.opentelemetry.io/otel` family to v1.44.0 to address three reachable CVEs (GO-2026-6061, GO-2026-5506, GO-2026-5426). The diff reviewed (the generated chunk) covers the semconv `metric.go` file rename from `v1.39.0/otelconv` to `v1.41.0/otelconv`. All changes are correct upstream vendor code:

- **`Enabled()` guard checks** added before `sync.Pool` allocations — correct performance optimization that short-circuits when the instrument is disabled.
- **`clear(*o)` before `*o = (*o)[:0]`** in deferred pool returns — correctly zeroes slice elements to release attribute references and prevent GC retention leaks when pooled slices hold `attribute.KeyValue` pointers.
- **New `*Observable` struct types** (`SDKExporterLogExportedObservable`, `SDKExporterSpanInflightObservable`, etc.) — consistent, correct additions following the established pattern in the file.
- **Semconv path rename** `v1.39.0/otelconv → v1.41.0/otelconv` — matches the version bump.

No issues found in the changed code. The PR description confirms `govulncheck ./...` returns 0 reachable vulnerabilities post-bump and `go build ./...` passes across all six modules.

joe0BAB (approved):

PR #610, opened 2026-08-06
feat(pass): add optional client timeout to run command

joe0BAB:
Add a RunOption to RunCommand so callers can forward a request timeout to client.New via client.WithTimeout. When the option is not provided, the client's default request timeout applies; a timeout of 0 disables the request timeout.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The implementation is clean and correct. Key points verified by the review:

- **Pointer-to-Duration for zero-value semantics**: `*time.Duration` correctly distinguishes "option not provided" (nil) from "timeout of 0" (disables the request timeout), matching the documented behavior.
- **No aliasing bug**: The option closure captures the duration value before taking its address — safe and idiomatic Go.
- **Correct nil guard**: `client.WithTimeout` is only appended when the option is non-nil, preserving the default client timeout when no option is passed.
- **`go.work.sum`**: Only checksum additions for transitive dependencies — no logic changes.

No bugs, logic errors, resource leaks, or security issues introduced by this PR.

Benehiko (approved):

PR #611, opened 2026-08-10
feat(pass): cancel run if the engine becomes unavailable mid-resolution

joe0BAB:
Start a background liveness monitor alongside resolveEnv that pings Version() every 5s (3s deadline per ping). If the engine stops responding the monitor cancels the shared context, unblocking pass run instead of hanging indefinitely.

The engine is multi-threaded and services pings concurrently with in-flight secret resolution and open ACM approval prompts, so a failed ping reliably indicates a wedge rather than load.

This mirrors the pattern docker/cli settled on after indefinite hangs were reported against an unreachable daemon:
- moby/moby#43376, moby/moby#28889: hang root causes
- docker/cli#2424: bounded preflight ping (2s) for TCP
- docker/cli#3722: extended the fix to unix sockets


docker-agent (commented):
### Assessment: 🟢 APPROVE

The liveness monitor implementation looks correct. The goroutine lifecycle is clean: `ctx.Done()` exits the loop on any cancellation (from either the parent context or `cancel(nil)` deferred by `RunCommand`), so there is no goroutine leak. The `context.WithCancelCause` / `context.Cause` pattern correctly surfaces the "secrets engine became unavailable" error when the monitor fires, and passes through genuine resolution errors when the monitor hasn't fired yet. The ping uses `context.WithoutCancel` to deliberately decouple the ping deadline from the already-cancelled parent — that's the right call. The test suite covers both the happy path and the failure path with appropriate intervals.

No bugs introduced by this PR were found.

plugins/pass/commands/run.go
```diff
@@ -182,6 +180,37 @@ func mergeEnv(processEnv, files []string) ([]string, error) {
...
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				pingCtx, pingCancel := context.WithTimeout(context.WithoutCancel(ctx), pingTimeout)
+				_, err := c.Version(pingCtx)
+				pingCancel()
+				if err != nil {
```
Benehiko: **[CONFIRMED] correctness — any `Version()` error cancels the run, not just engine death**

The monitor treats *any* `Version()` error as engine death, including non-availability failures. Two concrete cases:

- A daemon reporting a non-strict-semver version (`dev`, `1.2.3` without the `v`): `x/api/version.go` `NewVersion` rejects it and `client/client.go:153` wraps it as an error.
- An older daemon lacking the `healthv1` `VersionService`: connect returns `Unimplemented`, which `isDialError` does not match.

In both cases the first ping at t=5s fails and any resolution still in flight (e.g. an open approval prompt) is aborted with `secrets engine became unavailable: parsing daemon version ...` even though the engine is healthy and serving `GetSecrets`. Consider only cancelling on availability-class errors (dial failures / `client.ErrSecretsEngineNotAvailable`).

plugins/pass/commands/run.go
```diff
@@ -182,6 +180,37 @@ func mergeEnv(processEnv, files []string) ([]string, error) {
...
+	ctx, cancel := context.WithCancelCause(ctx)
+	go func() {
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				pingCtx, pingCancel := context.WithTimeout(context.WithoutCancel(ctx), pingTimeout)
```
Benehiko: **[CONFIRMED] correctness — a single slow ping kills the run, and the real header deadline is 1s, not 3s**

There is no retry or failure threshold, and the client is built without `WithResponseTimeout`, so `Transport.ResponseHeaderTimeout` stays at 1s (`x/api/defaults.go:33`, `client/client.go:215`) — capping the `GetVersion` header wait *below* the 3s `defaultLivenessPingTimeout`, which never comes into play for the header phase.

If the engine takes 1.2s to send `GetVersion` response headers (GC pause, busy host, laptop sleep/wake) while the user has an approval prompt open, the ping fails once with `timeout awaiting response headers` and cancel fires immediately (lines 204–207 give no second chance), aborting the run mid-approval. This contradicts the comment's own claim that "a failed ping means wedge, not load". Consider requiring N consecutive failures before cancelling.

plugins/pass/commands/run.go
```diff
@@ -98,27 +98,28 @@ func RunCommand(options ...RunOption) *cobra.Command {
 				return err
 			}
 
-			env, err := resolveEnv(cmd.Context(), c, merged)
+			ctx, cancel := startLivenessMonitor(cmd.Context(), c, defaultLivenessInterval, defaultLivenessPingTimeout)
+			defer cancel(nil)
+
+			env, err := resolveEnv(ctx, c, merged)
 			if err != nil {
+				if cause := context.Cause(ctx); cause != nil {
```
Benehiko: **[CONFIRMED] correctness — unconditional `context.Cause` preference masks real `resolveEnv` errors**

`if cause := context.Cause(ctx); cause != nil { return cause }` replaces `resolveEnv`'s error whenever the derived ctx is done for *any* reason:

1. When the consumer CLI wires the command context through signal handling, Ctrl-C mid-resolution returns the parent's cause — bare `context.Canceled`, or with this repo's `x/oshelper.NotifyContext` an `errCtxSignalTerminated` whose `Error()` returns `""` (`x/oshelper/context.go:27-29`), so cobra prints `Error: ` with an empty message — and the informative `resolving KEY: ...` error is discarded.
2. Race: `resolveEnv` fails with a genuine error (`2 secrets matched`, `ErrNotFound`) at the same moment a ping fails — the user is told `secrets engine became unavailable` and never sees the real misconfiguration.

Suggested fix: prefer the cause only when `errors.Is(err, context.Canceled)`.

Benehiko: **[CONFIRMED] test-coverage — the cause-preference branch (lines 106–108) is exercised by no test**

`TestStartLivenessMonitor` calls `startLivenessMonitor`+`resolveEnv` directly (never `RunE`), and both `TestRunCommand` wrapper subtests use envs without `se://` refs. A regression here — branch inverted, cause dropped, raw ctx error returned instead of the liveness cause — passes the full suite and silently loses the only actionable error message this feature adds.

plugins/pass/commands/run.go
```diff
@@ -98,27 +98,28 @@ func RunCommand(options ...RunOption) *cobra.Command {
 				return err
 			}
 
-			env, err := resolveEnv(cmd.Context(), c, merged)
+			ctx, cancel := startLivenessMonitor(cmd.Context(), c, defaultLivenessInterval, defaultLivenessPingTimeout)
+			defer cancel(nil)
```
Benehiko: **[CONFIRMED] correctness — `defer cancel(nil)` scopes the monitor to all of `RunE`, so it pings for the child process's entire lifetime**

Nothing consumes the derived ctx after `resolveEnv` returns (line 104), yet the monitor keeps pinging `Version()` every 5s until the child exits. For `pass run -- server` running for days that is ~17k pointless Version RPCs/day, a unix-socket connection pinned open (the 5s cadence defeats the 90s `IdleConnTimeout`), and engine log noise; if the engine legitimately restarts mid-run, the monitor cancels an unobserved ctx and exits. It also starts even when the env has zero `se://` refs, where the command previously never contacted the daemon.

This contradicts the commit's own "mid-resolution" scope (and subtest 3's name). Suggested fix: call `cancel(nil)` immediately after `resolveEnv` returns instead of deferring.

plugins/pass/commands/run.go
```diff
@@ -182,6 +180,37 @@ func mergeEnv(processEnv, files []string) ([]string, error) {
...
 
+// startLivenessMonitor returns a derived context that is cancelled if the
+// engine stops responding to Version() pings. The caller must defer cancel.
+// The engine handles pings concurrently with open approval prompts, so a
+// failed ping means wedge, not load.
+//
+// Pattern mirrors docker/cli#2424 and docker/cli#3722, which added a bounded
+// preflight ping to the Docker CLI after moby/moby#43376 and moby/moby#28889
+// showed that an unreachable daemon causes indefinite hangs.
+func startLivenessMonitor(ctx context.Context, c client.Client, interval, pingTimeout time.Duration) (context.Context, context.CancelCauseFunc) {
```
Benehiko: **[CONFIRMED] design — an out-of-band `Version()` ping cannot detect the per-request wedge, and liveness sits at the wrong layer**

If a provider plugin deadlocks or an approval prompt is lost while the engine process stays healthy, pings succeed every 5s forever, and with the default `requestTimeout=0` (`x/api/defaults.go:29`) `resolveEnv` hangs indefinitely — exactly the moby/moby#28889-class hang the comment cites. The headline failure mode therefore persists.

Meanwhile every other consumer of `github.com/docker/secrets-engine/client` (`ListPlugins`, external `GetSecrets` callers) gets no protection, because the mechanism is hand-wired into one `RunE` instead of offered as a client option. Consider moving liveness into the shared client module.

plugins/pass/commands/run.go
```diff
@@ -37,11 +37,13 @@ import (
 
 const sePrefix = "se://"
 
-// ExitCodeError is returned from RunCommand when the executed child process
-// terminated with a non-zero status. It carries the exit code the wrapper
-// should exit with. Returning this instead of calling os.Exit directly lets
-// the surrounding OTel span wrapper finish recording metrics and span data
-// before the process exits.
+const (
+	defaultLivenessInterval    = 5 * time.Second
```
Benehiko: **[CONFIRMED] design — liveness interval/ping-timeout are hardcoded with no `RunOption`, flag, or opt-out**

This breaks the configurability precedent `WithTimeout` set one commit earlier (6ff32c8) for the same embedder audience. Users on slow VMs or paused Docker Desktop who hit the false-positive cancellations flagged in the other comments have no lever — the only remediation is patching the package and re-vendoring. Consts named `default*` imply an override mechanism that does not exist.

plugins/pass/commands/run.go
```diff
@@ -182,6 +180,37 @@ func mergeEnv(processEnv, files []string) ([]string, error) {
 	return out, nil
 }
 
+// startLivenessMonitor returns a derived context that is cancelled if the
+// engine stops responding to Version() pings. The caller must defer cancel.
+// The engine handles pings concurrently with open approval prompts, so a
+// failed ping means wedge, not load.
+//
+// Pattern mirrors docker/cli#2424 and docker/cli#3722, which added a bounded
```
Benehiko: **[PLAUSIBLE] documentation — doc comment mis-cites precedent and promises behavior the code lacks**

docker/cli#2424 is lazy feature detection (it *removed* an eager ping), and none of the cited changes is a periodic poller that cancels in-flight work (verified via the GitHub API: #2424 defers pings; only #3722 loosely matches). The "preflight ping" framing is also false — `time.NewTicker` sends no ping until t=5s, so an engine wedged from the start is detected only at ~interval+ping ≈ 8s.

Marked plausible rather than confirmed because the commit scopes itself to *mid-resolution* cancellation, so the missing preflight is a doc/hardening gap rather than a clear defect — but the citation should be fixed so future maintainers don't defend a design upstream never shipped.

plugins/pass/commands/run.go
```diff
@@ -182,6 +180,37 @@ func mergeEnv(processEnv, files []string) ([]string, error) {
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				pingCtx, pingCancel := context.WithTimeout(context.WithoutCancel(ctx), pingTimeout)
+				_, err := c.Version(pingCtx)
+				pingCancel()
+				if err != nil {
+					cancel(fmt.Errorf("secrets engine became unavailable: %w", err))
```
Benehiko: **[CONFIRMED] error-handling — cause message double-wraps dial failures; availability sentinel inconsistent for non-dial ones**

`cancel` wraps `client.Version`'s error, which for dial errors already reads `secrets engine is not available: ...` (`client/client.go:147`). Socket removed mid-resolution yields: `secrets engine became unavailable: secrets engine is not available: dial unix ...: connect: no such file or directory` — the same fact twice.

Meanwhile header-timeout ping failures don't match `client.ErrSecretsEngineNotAvailable` at all, so a consumer classifying errors via `errors.Is(err, client.ErrSecretsEngineNotAvailable)` (friendly hint, exit-code mapping) catches the dial flavor but misses the timeout flavor of the same user-visible condition.

plugins/pass/commands/run.go
```diff
@@ -157,9 +158,6 @@ func RunCommand(options ...RunOption) *cobra.Command {
 	return cmd
 }
 
-// mergeEnv folds the process environment and any --env-file inputs into a
-// single deterministic KEY=VALUE slice. Precedence: process env first, then
```
Benehiko, line 161: **[CONFIRMED] documentation — unrelated doc-comment churn in this commit deletes documented contracts**

Several doc contracts are removed or distorted by this diff: `mergeEnv`'s precedence/determinism doc is gone entirely, `WithTimeout` lost its three-state (absent/0/>0) contract that motivates the `*time.Duration` pointer, `ExitCodeError` lost the "exit code the wrapper should exit with" caller contract, and `resolveRef`'s wildcard rationale is rewritten to say `ParseID` rejects wildcards before `ParsePattern` broadens the lookup — which misstates why the `ParseID` pre-check must not be removed as redundant.

The guarantee "process env first, then each file in order; later entries override" plus sorted-deterministic output now exists nowhere at function level, and `git blame` on all these lines now points at an unrelated liveness commit. Suggest restoring these docs or moving the churn to a separate commit.

plugins/pass/commands/run_test.go
```diff
@@ -319,6 +320,79 @@ func waitForReady(t *testing.T, r io.Reader) {
...
+}
+
+func (m *mockEngineClient) Version(ctx context.Context) (client.DaemonVersion, error) {
+	return m.versionFn(ctx)
+}
+
+func TestStartLivenessMonitor(t *testing.T) {
+	t.Parallel()
+
+	t.Run("resolves normally when engine stays alive", func(t *testing.T) {
```
Benehiko: **[CONFIRMED] test-coverage — the monitor's two core paths are untested: a hanging `Version()` and continue-after-successful-ping**

Every mock `versionFn` returns instantly and ignores its ctx, so dropping the `context.WithTimeout` on `pingCtx` (or a client change that stops honoring ctx in `Version`) passes the suite while a truly wedged engine blocks the ping forever and the monitor never fires — the exact hang class this PR exists for. A `versionFn` that blocks on `<-ctx.Done()` would cover it.

Subtest 1 ("resolves normally when engine stays alive", 50ms interval) resolves in microseconds and usually cancels before the first tick, making it functionally identical to subtest 3 — a regression breaking continue-after-successful-ping passes CI.

plugins/pass/commands/run_test.go
```diff
@@ -319,6 +320,79 @@ func waitForReady(t *testing.T, r io.Reader) {
...
+		require.NoError(t, err)
+		assert.Equal(t, []string{"X=val"}, out)
+	})
+
+	t.Run("cancels resolveEnv when engine becomes unavailable", func(t *testing.T) {
+		t.Parallel()
+		engineErr := errors.New("connection refused")
+		c := &mockEngineClient{
+			getSecretsFn: func(ctx context.Context, _ secrets.Pattern) ([]secrets.Envelope, error) {
+				<-ctx.Done()
```
Benehiko: **[CONFIRMED] test-quality — this subtest deadlocks instead of failing fast if the monitor regresses**

`getSecretsFn` blocks solely on `<-ctx.Done()`, whose only unblock path is the monitor's own cancel. If `startLivenessMonitor` regresses (ticker/cancel path broken), `resolveEnv` never returns; `t.Context()` cancels only after the test function returns, which never happens — the package hangs until the `go test` 10-minute panic with a goroutine dump instead of a clean assertion failure.

Guard with a deadline parent for the monitor, e.g. `context.WithTimeout(t.Context(), ...)`.

plugins/pass/commands/run_test.go
```diff
@@ -319,6 +320,79 @@ func waitForReady(t *testing.T, r io.Reader) {
...
+	})
+
+	t.Run("does not fire when resolution completes first", func(t *testing.T) {
+		t.Parallel()
+		c := &mockEngineClient{
+			getSecretsFn: func(_ context.Context, _ secrets.Pattern) ([]secrets.Envelope, error) {
+				return []secrets.Envelope{{ID: secrets.MustParseID("tok"), Value: []byte("val")}}, nil
+			},
+			versionFn: func(_ context.Context) (client.DaemonVersion, error) {
+				t.Error("Version() called unexpectedly")
```
Benehiko: **[CONFIRMED] test-quality — `t.Error` inside `versionFn` runs on the monitor goroutine and is safe only because `interval=time.Hour`**

If a future edit lowers the interval (e.g. copying subtest 1's 50ms "to actually exercise the path"), the ticker can fire after the subtest returns — and the in-flight ping is uncancellable because `pingCtx` derives from `context.WithoutCancel` (run.go:201) — so `t.Error` fires post-completion and the "Log in goroutine after test has completed" panic aborts the whole test binary.

The subtest name "does not fire when resolution completes first" also documents a contract production does not have: in `RunCommand` the monitor keeps running after resolution (see the `defer cancel` comment); the test passes only because an hour never elapses.

plugins/pass/commands/run_test.go
```diff
@@ -319,6 +320,79 @@ func waitForReady(t *testing.T, r io.Reader) {
 	go func() { _, _ = io.Copy(io.Discard, r) }()
 }
 
+type mockEngineClient struct {
```
Benehiko: **[CONFIRMED] simplification — `mockEngineClient` re-implements what `testhelper.MockResolver` (already used at line 106 of this file) provides**

The duplication exists because `startLivenessMonitor` demands a full `client.Client` while calling only `Version()`. The two `GetSecrets` mocks can drift: `MockResolver` does `pattern.Match` + sorting, while these hand-rolled closures return a static envelope ignoring the pattern, so pattern-dependent `resolveEnv` changes are silently unexercised by the liveness tests.

Narrowing the monitor's dependency to `ping func(context.Context) error` (or a one-method interface) deletes `mockEngineClient` entirely; only the blocking-`GetSecrets` subtest needs a custom resolver func.

Benehiko (commented):
## Code review — liveness monitor (ca71693)

15 findings from a multi-angle review with adversarial verification (14 CONFIRMED, 1 PLAUSIBLE, 0 refuted). 14 are posted as inline comments below; one finding has no diff anchor, so it is included here.

**Top risks:** false-positive run cancellation (any `Version()` error is treated as engine death, a single slow ping kills the run, and the effective header deadline is 1s rather than the intended 3s), error masking by the `context.Cause` branch, and the monitor both outliving resolution and being structurally unable to catch the per-request wedge it was built for.

---

**[CONFIRMED] correctness — `run.go:92` (not in diff): feature premise contradicts client config — missing `WithResponseTimeout(0)`**

`run` expects resolution to block for minutes on approval prompts, but builds the client without `WithResponseTimeout(0)`, leaving the 1s header timeout that `client.WithResponseTimeout`'s godoc says interactive flows must override.

If the engine does not flush response headers before blocking on approval, every approval taking longer than ~1s dies with a raw `timeout awaiting response headers` — before the monitor's first 5s tick — making the new mechanism and its curated error unreachable for the pre-header wedge class. If the engine does flush early, the unbounded post-header wait affects all client consumers and only `pass run` gets a (partial) guard. The timeout regimes (1s header / 3s ping / 5s interval / optional `--timeout`) live in different layers with no coordination.

---

Benehiko (approved):

PR #612, opened 2026-08-12
feat(accesscontrol): expose signature-bound bundle name on Darwin

joe0BAB:
Add BundleName to SigningIdentity: the human-readable application name from the Info.plist bound into the code signature (kSecCodeInfoPList, CFBundleDisplayName falling back to CFBundleName). Present for .app bundles and for bare binaries with an embedded __info_plist section; empty otherwise.

Display-only: the name is chosen freely by the signer, so it is not unique and must not be used as a trust key.


x/api/accesscontrol/accesscontrol_darwin.go
```diff
@@ -62,6 +62,13 @@ type SigningIdentity struct {
 	// so suitable for display/logging rather than as a sole trust key.
 	Organization string
 
+	// BundleName is the human-readable application name from the Info.plist
+	// bound into the code signature (kSecCodeInfoPList, CFBundleDisplayName
+	// falling back to CFBundleName), e.g. "Docker Desktop". Present for .app
+	// bundles and bare binaries with an embedded __info_plist section; empty
+	// otherwise. Display-only: chosen freely by the signer, not unique.
+	BundleName string
```
docker-agent, line 70: **[MEDIUM] BundleName is defined and wired for deserialization, but no producer populates it**

`signingIdentityFromProto` reads `BundleName` from the incoming proto message (`p.GetBundleName()`), but a repo-wide search finds no code anywhere in this repository that calls `SecCodeCopySigningInformation`, reads `kSecCodeInfoPList`, or extracts `CFBundleDisplayName`/`CFBundleName` to actually set `bundle_name` in the proto before it arrives here. There is also no `signingIdentityToProto` / `ToProto` producer function.

If the producer that populates `bundle_name` lives in a separate repository (e.g. the Docker Desktop client binary), this server-side change is correct and complete — the field will be populated once that client is updated. If the producer was intended to live in this PR (e.g. a CGo layer calling the Security framework), it is missing and `BundleName` will always be empty.

Suggestion: confirm the producer is tracked in a companion PR/issue, or add a brief code comment on the field noting which component is responsible for populating it, so reviewers can verify end-to-end completeness.

docker-agent (commented):

One potential issue was found in the changes introduced by this PR.

Benehiko (approved):

PR #619, opened 2026-08-18
feat(accesscontrol): accept multiple patterns per access check

joe0BAB:
CheckAccessRequest now carries a list of patterns instead of a single embedded pattern, so one access decision (and one approval prompt) can cover every pattern of a request. The proto message keeps field 1 and becomes repeated; the service parses each pattern and rejects requests without any.


docker-agent (commented):

Two medium-severity findings in the new multi-pattern access control change — one proto wire-compatibility concern and one validation gap introduced by the PR. No high-severity bugs found.

x/api/accesscontrol/accesscontrol.go
```diff
@@ -25,7 +25,7 @@ type AccessControl interface {
 }
 
 type CheckAccessRequest struct {
-	secrets.Pattern
+	Patterns []secrets.Pattern
```
docker-agent, line 28: **[MEDIUM] Direct callers of `CheckAccess` can pass an empty `Patterns` slice, bypassing service-layer validation**

`CheckAccessRequest.Patterns` is now a `[]secrets.Pattern` slice, but the only guard against an empty slice (`len(msgPatterns) == 0`) lives at the RPC boundary in `service.go`. Any code that constructs and passes a `CheckAccessRequest` directly — without going through the connect handler — can call `ac.CheckAccess(ctx, CheckAccessRequest{})` or `CheckAccessRequest{Patterns: nil}` and reach the underlying implementation with zero patterns.

Before this PR, embedding a single `secrets.Pattern` made a zero-value struct produce one (potentially invalid) pattern; the new slice type makes the empty case silent and distinct. If the concrete `AccessControl` implementation iterates over `Patterns` with no nil/empty guard, this could cause incorrect access decisions or a panic.

**Suggestion**: Add a guard directly in the service's `CheckAccess` call path, or document that `AccessControl` implementations must handle empty `Patterns` explicitly. The most robust fix is to validate in `CheckAccessRequest` itself or at the interface boundary rather than only at the RPC layer.

x/api/accesscontrol/v1/api.proto
```diff
@@ -21,88 +21,72 @@ import "google/protobuf/timestamp.proto";
...
+  // Decides whether the requester may resolve the queried secrets.
+  // Called for every GetSecrets request.
   rpc CheckAccess(CheckAccessRequest) returns (CheckAccessResponse);
 }
 
 message CheckAccessRequest {
-  // What the caller is trying to resolve — mirrors the resolver pattern.
-  string pattern = 1;
+  // Patterns the caller wants to resolve; at least one required.
+  repeated string patterns = 1;
```
docker-agent, line 31: **[MEDIUM] Reusing field number 1 (`optional string` → `repeated string`) is a breaking wire-format change for mixed-version deployments**

Both `optional string` and `repeated string` use proto3 wire type 2 (length-delimited), so the field number is reused safely in one direction: an old client that sends a single `pattern` value will be decoded as a one-element `patterns` slice by the new server — correct. The problem runs the other way: a **new client** that sends multiple `patterns` values (field 1, repeated) to an **old server** will have the old server decode field 1 as an `optional string`, keeping only the **last** occurrence. During a rolling deployment where old and new server instances coexist, old servers will silently evaluate access based on only one of the patterns, potentially granting or denying access incorrectly for multi-pattern requests.

The PR description notes that "the proto message keeps field 1 and becomes repeated" but does not address this mixed-version hazard.

**Suggestion**: If simultaneous old/new server deployments are possible, consider either (a) assigning a new field number for `patterns` and deprecating field 1, or (b) documenting that the server must be fully rolled out before clients start sending multiple patterns. A field alias or a migration field (field 1 = single legacy pattern, field 3 = repeated patterns) would preserve full backward compatibility.

Benehiko (approved):

PR #621, opened 2026-08-19
feat(api): add Authorize preflight to the ACM API and the client

joe0BAB:
AccessControl gains Authorize: a preflight that grants the requester access to a set of patterns, valid until the returned expiry. The ACM wire API carries the patterns plus the engine-provided requester identity.

Clients preflight through the new secrets.Authorizer interface, backed by an AuthorizerService served by the engine next to ResolverService (and deliberately not by secrets-provider plugins). Only patterns go over the wire; the engine derives the requester identity from the connection peer.


x/api/accesscontrol/service.go
```diff
@@ -78,3 +72,48 @@ func (s *accessControlService) CheckAccess(ctx context.Context, c *connect.Reque
...
+			PID:                int(requester.GetPid()),
+			Name:               requester.GetName(),
+			AbsoluteBinaryPath: requester.GetAbsoluteBinaryPath(),
+		},
+		SigningInfo: signingInfoFromProto(requester),
+	}
+
+	expiry, err := s.ac.Authorize(ctx, req)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("authorize failed for %q: %w", msgPatterns, err))
```
docker-agent, line 103: **[MEDIUM] All errors from `ac.Authorize` are wrapped as `CodeInternal`, masking domain-level denial codes**

This PR introduces the `Authorize` method that unconditionally wraps every error from `s.ac.Authorize()` as `connect.CodeInternal`:

```go
expiry, err := s.ac.Authorize(ctx, req)
if err != nil {
    return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("authorize failed for %q: %w", msgPatterns, err))
}
```

If the access-control implementation returns a connect-typed error (e.g., `CodePermissionDenied` from a downstream call), `connect.NewError(connect.CodeInternal, ...)` replaces the original code entirely. Callers therefore cannot distinguish a transient server failure from an authorization denial — both surface as `Internal`. This makes it harder for clients to implement correct retry/backoff logic and prevents policy-driven "denied" responses from being surfaced accurately.

**Suggestion:** Inspect the error before wrapping — if it is already a `*connect.Error`, return it directly (or re-code only domain errors):
```go
if err != nil {
    var connectErr *connect.Error
    if errors.As(err, &connectErr) {
        return nil, connectErr
    }
    return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("authorize failed for %q: %w", msgPatterns, err))
}
```

x/api/resolver/resolver.go
```diff
@@ -115,3 +116,63 @@ func (r resolverClient) GetSecrets(ctx context.Context, pattern secrets.Pattern)
...
+		pattern, err := secrets.ParsePattern(msgPattern)
+		if err != nil {
+			return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("invalid pattern %q: %w", msgPattern, err))
+		}
+		patterns = append(patterns, pattern)
+	}
+
+	expiry, err := s.authorizer.Authorize(ctx, patterns...)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("authorize failed for %q: %w", msgPatterns, err))
```
docker-agent, line 152: **[MEDIUM] Errors from `authorizer.Authorize` are unconditionally wrapped as `CodeInternal`, losing the original error code**

This PR introduces the `authorizerService.Authorize` handler with the same error-wrapping pattern:

```go
expiry, err := s.authorizer.Authorize(ctx, patterns...)
if err != nil {
    return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("authorize failed for %q: %w", msgPatterns, err))
}
```

If `s.authorizer.Authorize` (a `secrets.Authorizer`) internally calls a connect RPC and propagates its error (e.g., `CodePermissionDenied`), wrapping it here replaces the original code with `CodeInternal`. The client calling this service receives an opaque internal error instead of an actionable denial code.

**Suggestion:** Pass through existing connect errors unchanged:
```go
if err != nil {
    var connectErr *connect.Error
    if errors.As(err, &connectErr) {
        return nil, connectErr
    }
    return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("authorize failed for %q: %w", msgPatterns, err))
}
```

docker-agent (commented):

This PR cleanly adds the `Authorize` preflight across the secrets engine stack — the interface definition, the engine-side handler, the ACM wire API handler, and the client adapter. The proto changes, generated code, and connect-go wiring all look correct. Two medium-severity issues were found in the new `Authorize` error-handling paths introduced by this PR.

Benehiko (approved):

PR #623, opened 2026-08-19
feat(accesscontrol): check access for a single pattern, authorize many

joe0BAB:
CheckAccess narrows back to one embedded pattern per request — it gates a single GetSecrets query. Authorize keeps the pattern list: one preflight can cover a whole set, and a denial maps to PermissionDenied via secrets.ErrAccessDenied on both the ACM and resolver surfaces.


Benehiko (approved):

x/api/accesscontrol/service.go
```diff
@@ -62,7 +62,7 @@ func (s *accessControlService) CheckAccess(ctx context.Context, c *connect.Reque
 
 	allowed, err := s.ac.CheckAccess(ctx, req)
```
docker-agent, line 63: **[LOW] CheckAccess handler does not map `ErrAccessDenied` → `PermissionDenied`; inconsistent with `Authorize`**

The `Authorize` handler explicitly checks `errors.Is(err, secrets.ErrAccessDenied)` and returns `CodePermissionDenied`, but `CheckAccess` wraps all errors — including any that could carry `ErrAccessDenied` — into `CodeInternal`:

```go
allowed, err := s.ac.CheckAccess(ctx, req)
if err != nil {
    return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("access check failed for %q: %w", msgPattern, err))
}
```

The `AccessControl` interface defines `CheckAccess` as returning `(bool, error)`, so denial is communicated via the boolean — an implementation *should* never return `ErrAccessDenied` as an error. However, there is no runtime guard preventing it, and if any future or third-party implementation returns `(false, ErrAccessDenied)`, the caller would receive a confusing `CodeInternal` instead of `CodePermissionDenied`.

Consider adding a guard to make the contract explicit and symmetric with `Authorize`:

```go
if err != nil {
    if errors.Is(err, secrets.ErrAccessDenied) {
        return nil, connect.NewError(connect.CodePermissionDenied, secrets.ErrAccessDenied)
    }
    return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("access check failed for %q: %w", msgPattern, err))
}
```

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly narrows `CheckAccessRequest` from a repeated patterns field to a single scalar pattern, updating proto, generated pb.go, Go interface structs, the service handler, and tests consistently. The `Authorize` RPC retains the multi-pattern list and correctly maps `ErrAccessDenied → PermissionDenied`. Proto/pb.go contract is sound — edition 2023 generates presence-tracking accessors (`HasPattern`, `ClearPattern`) for all singular fields by default, so no `optional` keyword is needed. One low-severity observation flagged inline.

x/api/accesscontrol/v1/api.proto
```diff
@@ -31,8 +31,8 @@ service AccessControlService {
 }
 
 message CheckAccessRequest {
-  // Patterns the caller wants to resolve; at least one required.
-  repeated string patterns = 1;
+  // Pattern the caller wants to resolve.
+  string pattern = 1;
```
docker-agent, line 35: **[MEDIUM] Proto field reuse: `patterns` (repeated string, field 1) replaced by `pattern` (singular string, field 1) — backward-incompatible wire change**

Both `repeated string` and `string` use wire type 2 (length-delimited), so the proto decoder will not reject the old format. However, the semantics change in a breaking way: old clients that send multiple patterns in `CheckAccessRequest` will have each element decoded as successive values for field 1; for a singular field in proto3/edition 2023, the **last value wins**, so all patterns except the last are silently dropped.

Any in-flight or not-yet-updated client that populates `patterns = ["foo/*", "bar/*"]` will end up with only `"bar/*"` being evaluated — a silent semantic failure with no error surfaced.

If this is an intentional breaking change (i.e., all callers are being updated atomically), consider reserving field number 1 with `reserved 1; reserved "patterns";` to protect against future accidental reuse and to document the break clearly. Otherwise, use a new field number for `pattern`.

x/api/accesscontrol/service.go
```diff
@@ -62,7 +62,7 @@ func (s *accessControlService) CheckAccess(ctx context.Context, c *connect.Reque
...
 	if err != nil {
-		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("access check failed for %q: %w", msgPatterns, err))
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("access check failed for %q: %w", msgPattern, err))
```
docker-agent, line 65: **[LOW] `CheckAccess` maps `ErrAccessDenied` (if returned as error) to `CodeInternal` — inconsistent with the `Authorize` path**

The `Authorize` handler includes an explicit guard:
```go
if errors.Is(err, secrets.ErrAccessDenied) {
    return nil, connect.NewError(connect.CodePermissionDenied, secrets.ErrAccessDenied)
}
```

The `CheckAccess` handler has no equivalent guard — any error from `s.ac.CheckAccess`, including a hypothetical `ErrAccessDenied`, is unconditionally surfaced as `CodeInternal`.

The design intent is likely that `CheckAccess` implementations must signal denial via `(false, nil)` rather than an error return, but this contract is not documented in the interface. If an implementation mistakenly returns `ErrAccessDenied` as the error value, callers see an unexpected internal error instead of a permission-denied status.

**Suggestion**: Either add the guard here as well:
```go
if errors.Is(err, secrets.ErrAccessDenied) {
    // Treat as denial rather than internal error
    return connect.NewResponse(accesscontrolv1.CheckAccessResponse_builder{
        Decision: &deny,
    }.Build()), nil
}
```
or add a comment to the `AccessControl` interface documenting that `CheckAccess` must never return `ErrAccessDenied` as an error.

docker-agent (commented):

This PR cleanly narrows `CheckAccess` to a single embedded pattern and introduces the new `Authorize` multi-pattern preflight. The logic split is sensible and test coverage for the new path is solid. Two findings need attention: a backward-incompatible proto wire change (same field number, repeated→singular) and a missing `ErrAccessDenied` guard in the `CheckAccess` error path.

PR #626, opened 2026-08-20
feat(accesscontrol): encode requester identity into canonical wire form

joe0BAB:
Add NewRequester, which builds the accesscontrolv1.Requester proto from ProcessInfo and platform SigningInfo as the inverse of signingInfoFromProto, with round-trip tests for each platform.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The `NewRequester` implementation correctly encodes `ProcessInfo` and platform-specific `SigningInfo` into the proto wire form as the inverse of `signingInfoFromProto`. Field mappings are consistent with the decoder side across all three platforms (Darwin, Linux, Windows), and the round-trip tests cover the key cases including empty/zero values, sentinel PID=-1, and platform-specific signing identity fields.

No bugs were found in the changed code.

Benehiko (approved):

PR #628, opened 2026-08-20
feat(realms): expose canonical realm sets

joe0BAB:
Add All and per-group accessors (AllAuth, AllMCP, AllSandbox, AllSecretsEngine) so consumers can use the canonical set of Docker realms instead of mirroring the package's realm vars. The unexported group slices are the static source of truth: All is their concatenation, and accessors return defensive copies.

Tests guard the set against drift: every exported realm var must be listed in exactly one group, and accessors must return copies.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes look correct. The canonical realm-set accessors (`All`, `AllAuth`, `AllMCP`, `AllSandbox`, `AllSecretsEngine`) all use `slices.Clone` to return defensive copies, preventing callers from mutating the internal backing slices. The `all` package-level var is correctly initialized via `slices.Concat` of the four group slices at package init time. All 13 realm patterns (6 auth + 3 MCP + 2 sandbox + 2 secrets-engine) are accounted for across the groups.

The test suite is well-structured: `TestAccessorsReturnCopies` correctly verifies copy semantics by checking whether the third `accessor()` call is affected by mutation of a prior return value, `TestRealmsListedOnce` guards against double-listing, and `TestEveryExportedRealmIsListed` guards against drift. No bugs were found in the changed code.

PR #629, opened 2026-08-20
feat(secrets): add Dedupe to remove subsumed patterns

joe0BAB:
Dedupe collapses a pattern list to its maximal elements: patterns whose matches are fully covered by another entry are dropped, equivalent spellings (e.g. **, **/* and */**) keep their first occurrence, and input order is preserved.

Containment is decided as true language containment over identifiers, computed on canonical token sequences with an O(L^2) DP per pair; length, shape and literal-anchor filters skip most pairs. This is deliberately stricter than Pattern.Includes, which also treats a literal as covering '*' and '*' as covering '**'.

Verified against a brute-force oracle: every pattern pair with up to three components, checked against all identifiers with up to nine components over a three-symbol alphabet.


docker-agent (commented):
### Assessment: 🟢 APPROVE

No bugs found in the changed code. The `Dedupe` function, `canonicalize`, and `includes` DP are all correct.

**Summary**: This PR introduces a well-implemented pattern deduplication algorithm. Key areas checked:

- **`canonicalize`**: Correctly collapses adjacent wildcards — `stars` and `gap` accumulate independently and are emitted together in `flush()`. Stars before `**` are **not** lost; they stay in `stars` and are emitted first. Patterns like `**/*`, `*/**`, and `**` all correctly collapse to the same `[*, **]` canonical form.
- **`includes` DP**: The two-row rolling array (`prev`/`cur`) is fully overwritten on every outer iteration (both `cur[nq]` at the top and all `cur[j]` in the inner loop), so no stale values bleed through. The `q[j]=tokenGap` case formula `p[i].kind==tokenStar && cur[j+1] && prev[j]` correctly requires both: gap=0 expansion (`cur[j+1]`) and gap≥1 expansion (`prev[j]`), and correctly rejects literals since they can't cover all gap expansions.
- **`Dedupe` loop**: The mutual-inclusion tie-breaking (keep earlier entry when `j < i`, keep `i` when mutual) is sound. Transitivity ensures that even if a dominator `j` is itself later dominated, `i` is still correctly dropped.
- **`subsumes` filters**: `firstLit`/`lastLit` are correctly `""` for wildcard-anchored entries (zero value of `string`), so the literal-anchor filters only fire when `other` truly starts/ends with a literal.
- **Tests**: The exhaustive oracle in `TestDedupeExhaustive` (checking all 84 patterns against all identifiers up to 9 components) provides strong correctness evidence.

PR #631, opened 2026-08-21
feat: return decision and expiry from Authorize

joe0BAB:
Authorize previously signalled denial only through an error and returned a bare expiry. It now returns an AuthorizeResponse value carrying the decision and its expiry, so a denial can be a valid, cacheable outcome instead of an error.

- add secrets.AuthorizeResponse{Expiry, Allow}; secrets.Authorizer and accesscontrol.AccessControl return it by value (zero value denies)
- add a Decision field to both AuthorizeResponse protos, introduce the Decision enum in resolver.v1, and regenerate
- map Allow to DECISION_ALLOW/DENY in the connect handlers and back in the clients; an unset decision on the wire is treated as deny
- keep mapping ErrAccessDenied to CodePermissionDenied
- expose the AuthorizeResponse alias in the client package


docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR correctly implements the `AuthorizeResponse` refactor across all layers. Full diff reviewed (797 lines across 13 files).

**Correctness properties verified:**

- **Deny-by-default holds at every layer**: `secrets.AuthorizeResponse{}` zero value has `Allow=false`; the proto `DECISION_DENY=0` is the wire zero value; `GetDecision()` returns `DECISION_DENY` for an absent field. All three agree.
- **Decision enum mapping is consistent**: `Allow=true → DECISION_ALLOW`, `false → DECISION_DENY` in both the resolver and accesscontrol service handlers; the inverse mapping on the client side is correct.
- **Unset-decision-on-wire → deny**: `GetDecision() == DECISION_ALLOW` is the only path that sets `Allow=true` on the client, so a missing field correctly defaults to deny.
- **`ErrAccessDenied` → `CodePermissionDenied`** mapping is preserved in both handlers.
- **Expiry propagation**: `ExpiresAt` is set in both handlers and recovered via `GetExpiresAt().AsTime()` on the client.
- **Proto/pb.go consistency**: The `Decision` field is marked optional (presence-tracked), allowing a missing field to be distinguished from `DECISION_DENY=0` — both correctly result in deny.
- **`client.AuthorizeResponse` alias** is correctly exposed.

No bugs found in the introduced code.

x/api/resolver/v1/api.proto
```diff
@@ -41,9 +41,18 @@ message AuthorizeRequest {
...
+}
+
 message AuthorizeResponse {
-  // When the authorization expires.
-  google.protobuf.Timestamp expires_at = 1;
+  // The authorization decision.
+  Decision decision = 1;
+
+  // When the authorization decision expires.
+  google.protobuf.Timestamp expires_at = 2;
```
docker-agent, line 55: **[HIGH] Breaking wire-format change: `expires_at` renumbered from field 1 → 2, new `decision` reuses field 1**

`AuthorizeResponse.expires_at` was previously field number 1. This PR renumbers it to field 2 and places the new `Decision decision` enum at field 1. Proto field numbers are permanently part of the wire encoding — this is a breaking change.

**What goes wrong in a rolling deployment:**
- An **old server** encodes `expires_at` as field 1 with wire type 2 (length-delimited, for `google.protobuf.Timestamp`).
- A **new client** expects field 1 to be the `decision` enum (wire type 0 / varint).
- When the Go proto runtime encounters a wire-type mismatch it silently discards the field per proto unknown-field rules.
- Result: `GetDecision()` returns the zero default (`DECISION_DENY`) even when the old server intended to allow — **all requests appear denied during the upgrade window**. The `expires_at` value is also silently dropped.

**Recommendation:** Add the `decision` field at a new, unused field number (e.g. field 3) and keep `expires_at` at its original field 1. Reusing field 1 for a different type is a proto anti-pattern that cannot be safely rolled out.

x/api/accesscontrol/v1/api.proto
```diff
@@ -45,8 +45,11 @@ message AuthorizeRequest {
 }
 
 message AuthorizeResponse {
-  // When the authorization expires.
-  google.protobuf.Timestamp expires_at = 1;
+  // The authorization decision.
+  Decision decision = 1;
+
+  // When the authorization decision expires.
+  google.protobuf.Timestamp expires_at = 2;
```
docker-agent, line 52: **[HIGH] Breaking wire-format change: `expires_at` renumbered from field 1 → 2 in accesscontrol proto**

Same proto wire-format regression as in `resolver.v1`. The `AuthorizeResponse.expires_at` field was at field number 1; this PR moves it to field 2 and assigns field 1 to the new `Decision decision` enum.

**Impact in a mixed-version environment:**
- Old servers encode `expires_at` as tag 1 / wire type 2 (Timestamp / length-delimited).
- New clients read tag 1 as `decision` (wire type 0 / varint). Type mismatch → field silently discarded.
- `GetDecision()` returns `DECISION_DENY` (zero default) for every response from old servers, regardless of what the server intended.
- Any access-control service deployment that cannot do a hard cutover (replace all servers before any client) will silently deny all `Authorize` calls during the upgrade window.

**Recommendation:** Assign the new `decision` field a number that has never been used in this message (e.g. field 3), preserving `expires_at` at field 1.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR correctly refactors the Authorize API to return `AuthorizeResponse{Expiry, Allow}` instead of a bare `time.Time`, and the zero-value-is-deny property is sound. However, there are **two high-severity proto wire-format breaking changes** and two medium-severity issues around zero-time serialization that should be addressed before merging.

x/api/resolver/resolver.go
```diff
@@ -144,15 +143,20 @@ func (s authorizerService) Authorize(ctx context.Context, c *connect.Request[res
...
 		}
 		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("authorize failed for %q: %w", msgPatterns, err))
 	}
+	decision := resolverv1.Decision_DECISION_DENY
+	if resp.Allow {
+		decision = resolverv1.Decision_DECISION_ALLOW
+	}
 	return connect.NewResponse(resolverv1.AuthorizeResponse_builder{
-		ExpiresAt: timestamppb.New(expiry),
+		ExpiresAt: timestamppb.New(resp.Expiry),
```
docker-agent, line 158: **[MEDIUM] Zero `time.Time{}` serialized as a large-negative Timestamp — undefined semantics for DENY responses**

`timestamppb.New(resp.Expiry)` is called unconditionally. When `resp.Expiry` is the zero `time.Time{}` (e.g. for a DENY decision, where the `AuthorizeResponse` zero value has `Expiry: time.Time{}`), this serializes Go's zero time — January 1, year 1 — as a Timestamp with a very large negative `Seconds` field (~`-62135596800`).

The proto contract does not document what `expires_at` means when `decision = DECISION_DENY`. Any consumer (caching layer, retry scheduler, TTL manager) that reads `ExpiresAt` without first checking `Decision` will see a bizarre value in the distant past. Depending on interpretation:
- "already expired" → re-authorize immediately on every call (cache miss storm)
- treated as a sentinel for "no expiry" → cache the DENY forever

**Recommendation:** Omit `ExpiresAt` when the decision is DENY or when `resp.Expiry.IsZero()`, so consumers see a nil/absent field rather than an ambiguous year-1 timestamp:
```go
var expiresAt *timestamppb.Timestamp
if !resp.Expiry.IsZero() {
    expiresAt = timestamppb.New(resp.Expiry)
}
return connect.NewResponse(resolverv1.AuthorizeResponse_builder{
    ExpiresAt: expiresAt,
    Decision:  &decision,
}.Build()), nil
```

x/api/accesscontrol/service.go
```diff
@@ -95,15 +95,20 @@ func (s *accessControlService) Authorize(ctx context.Context, c *connect.Request
...
 		}
 		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("authorize failed for %q: %w", msgPatterns, err))
 	}
+	decision := accesscontrolv1.Decision_DECISION_DENY
+	if resp.Allow {
+		decision = accesscontrolv1.Decision_DECISION_ALLOW
+	}
 	return connect.NewResponse(accesscontrolv1.AuthorizeResponse_builder{
-		ExpiresAt: timestamppb.New(expiry),
+		ExpiresAt: timestamppb.New(resp.Expiry),
```
docker-agent, line 110: **[MEDIUM] Zero `time.Time{}` serialized as a large-negative Timestamp in accesscontrol service — undefined semantics for DENY**

Same issue as in `resolver.go`: `timestamppb.New(resp.Expiry)` is called unconditionally, and when `resp.Expiry` is the zero `time.Time{}` (DENY decision with no expiry), it produces a Timestamp with `Seconds ≈ -62135596800` (Go's year-1 zero time, not Unix epoch).

Any consumer of the access-control `AuthorizeResponse.expires_at` that doesn't check `decision` first will receive a semantically undefined value. Consider guarding with `resp.Expiry.IsZero()`:
```go
var expiresAt *timestamppb.Timestamp
if !resp.Expiry.IsZero() {
    expiresAt = timestamppb.New(resp.Expiry)
}
return connect.NewResponse(accesscontrolv1.AuthorizeResponse_builder{
    ExpiresAt: expiresAt,
    Decision:  &decision,
}.Build()), nil
```

Benehiko (approved):

PR #640, opened 2026-09-08
fix(pass): fail with a non-zero exit code when removing a missing secret

joe0BAB:
`docker pass rm <name>` used to print `RM: <name>` and exit 0 even when no such secret existed, claiming a removal that never happened. Every sibling `docker ... rm` (volume, network, context, buildx) errors and exits non-zero on a missing target.

rm now checks which secrets exist (via GetAllMetadata) before deleting: missing names print `ERR: <name>: secret not found` and the command exits non-zero, while existing names are still removed. The store's Delete stays deliberately idempotent since Upsert relies on it.

Note: `docker mcp secret rm` shells out to `docker pass rm` and will surface this error for missing secrets; mcp-gateway gets adjusted independently.


docker-agent (commented):
### Assessment: 🟢 APPROVE

Benehiko (approved):

PR #641, opened 2026-09-08
feat(pass): explain how to recover from a locked keychain collection

Benehiko:
`store` v0.4.0 classifies a locked Secret Service collection as `keychain.ErrCollectionLocked`. This adds the CLI-side half: the store-backed commands (`set`, `get`, `ls`, `rm`) append a one-line recovery hint to any error matching it, so the user gets the remediation instead of an opaque failure.

```
Error: get 'foo': keychain collection is locked: ...

Unlock the keychain and retry: log in to the desktop session, or run gnome-keyring-daemon --unlock on a headless host.
```

A locked collection is the default state on headless Linux hosts with SSH key-only logins, where PAM has no password to auto-unlock the login keyring. Per the keychain README, the hint is the prescribed handling: the secrets still exist in the locked collection, so the CLI tells the user how to unlock it and never falls back to another store.

Each command constructor applies the wrapper itself, so a CLI that mounts them needs no extra wiring. Other errors pass through unchanged, and `errors.Is(err, keychain.ErrCollectionLocked)` still matches the wrapped error.

## Testing

Unit tests cover the wrapper directly (hint on locked errors from `PreRunE` and `RunE`, pass-through for other errors, nil, hook-less commands) and end to end through `set`, `get`, `ls`, and `rm` against a mock store that returns a locked error. Live locked-collection behavior is already covered by the store's fake-seam tests and the `ubuntu-24-gnome-keyring-locked` CI target.


docker-agent (commented):
### Assessment: 🟢 APPROVE

joe0BAB (approved):

PR #644, opened 2026-09-10
fix: replace Pattern.Includes with Contains and Overlaps

joe0BAB:
`Pattern.Includes` reused the ID matcher and treated a wildcard in the candidate pattern as matching anything, so it implemented neither containment nor overlap: it reported overlapping-but-not-contained patterns as included (`docker/proj1/**` claimed to include `docker/*/mcp/*`), was asymmetric, and gave different answers for spellings with equal match sets (`a/a/**` vs `a/a/**/**`).

Two changes:

**Replace `Includes` with `Contains` and `Overlaps`.** Each relation is decided by a two-row DP over canonicalized wildcard tokens (runs of `*`/`**` collapse; a lone `**` becomes `*/**` since IDs are never empty):

- `p.Contains(other)` — p matches every ID that other matches: the containment Includes documented all along, with the same operand order.
- `p.Overlaps(other)` — p and other match at least one ID in common: symmetric, and spelling-insensitive by construction.

Both run in `O(n*m)` time and were verified against a brute-force set oracle on all 115,600 pattern pairs of up to 4 components. The ID matcher drops the candidate-side `*` special case it only carried for Includes.

**Remove `Filter`.** It reduced two patterns to the contained one and returned false for overlapping-but-not-contained pairs, silently dropping providers whose namespace merely overlaps a query. No signature-compatible definition is exact, because the intersection of two patterns is not always expressible as a single pattern; it is removed until a caller needs a well-defined replacement.


docker-agent (commented):
### Assessment: 🟢 APPROVE

x/secrets/pattern.go
```diff
@@ -77,15 +71,32 @@ func isValidPatternRune(c rune) bool {
...
-	Includes(other Pattern) bool
-	// String formats the [Pattern] as a string
+	// Contains reports whether the pattern matches every ID [other]
+	// matches:
+	//
+	//	p.Contains(other)  ⇔  matches(other) ⊆ matches(p)
+	//
+	// Containment implies overlap. O(n*m) time, O(m) space, for component
+	// counts n and m.
+	Contains(other Pattern) bool
```
Benehiko, line 93: Can we have an example pattern to show the outcome here?

x/secrets/pattern.go
```diff
@@ -101,35 +112,139 @@ func (p pattern) Match(id ID) bool {
...
+	tokenGap
+)
+
+type token struct {
+	kind tokenKind
+	lit  string
+}
+
+// canonicalize tokenizes a pattern so that two patterns match the same IDs
+// iff their tokens are equal: a wildcard run collapses to its '*'s plus one
```
Benehiko:
```suggestion
// if their tokens are equal: a wildcard run collapses to its '*' plus one
```

x/secrets/pattern.go
```diff
@@ -101,35 +112,139 @@ func (p pattern) Match(id ID) bool {
...
+// trailing '**' if it had any, and a pure '**' becomes "*/**" since IDs are
```
Benehiko:
```suggestion
// trailing '**' if it had any, and a standalone '**' becomes "*/**" since IDs are
```
pure?

actually no idea what the sentences here mean

x/secrets/pattern.go
```diff
@@ -101,35 +112,139 @@ func (p pattern) Match(id ID) bool {
...
+	}
+	flush()
+	if len(toks) == 1 && toks[0].kind == tokenGap {
+		toks = []token{{kind: tokenStar}, {kind: tokenGap}}
+	}
+	return toks
+}
+
+// covers reports whether p matches every ID q matches, for canonical tokens.
+// The component alphabet is unbounded, so a literal in p never covers a '*'
```
Benehiko:
```suggestion
// The component alphabet is unbounded, so a literal in p never covers an '*'
```

x/secrets/pattern_test.go
```diff
@@ -90,17 +90,61 @@ func TestPatternIncludes(t *testing.T) {
...
+		// docker/x/mcp/y matches the other pattern but not this one.
+		{"docker/proj1/**", "docker/*/mcp/*", false},
 		{"docker/proj1/**", "docker/**/mcp/**", false},
 		{"docker/**", "docker/**/mcp/**", true},
+		// ⟦a/a/**⟧ == ⟦a/a/**/**⟧: equivalent spellings must behave
+		// identically (the old Includes was spelling-sensitive and
+		// answered true for a/a/** but false for a/a/**/**).
+		{"a/a/**", "a/a/**/**", true},
+		{"a/a/**/**", "a/a/**", true},
+		{"a/a/*", "a/a/**", false},
```
Benehiko, line 101: and `{"a/a/**", "a/a/*", true}`?

x/secrets/pattern_test.go
```diff
@@ -90,17 +90,63 @@ func TestPatternIncludes(t *testing.T) {
...
 		{"*", "*/foo", false},
 		{"docker/*/mcp/*", "docker/proj1/**", false},
-		{"docker/proj1/**", "docker/*/mcp/*", true},
+		// docker/x/mcp/y matches the other pattern but not this one.
+		{"docker/proj1/**", "docker/*/mcp/*", false},
 		{"docker/proj1/**", "docker/**/mcp/**", false},
 		{"docker/**", "docker/**/mcp/**", true},
+		// ⟦a/a/**⟧ == ⟦a/a/**/**⟧: equivalent spellings must behave
+		// identically (the old Includes was spelling-sensitive and
+		// answered true for a/a/** but false for a/a/**/**).
```
Benehiko: encoding history

x/secrets/pattern.go
```diff
@@ -101,35 +113,148 @@ func (p pattern) Match(id ID) bool {
...
+		}
+		prev, cur = cur, prev
+	}
+	return prev[0]
+}
+
+func compatible(p, q []token) bool {
+	np, nq := len(p), len(q)
+	// dp[j] == "p[i:] and q[j:] match a common suffix"; row i needs only row
+	// i+1 (prev).
```
Benehiko: this comment made no sense to me.

x/secrets/pattern.go
```diff
@@ -101,35 +113,148 @@ func (p pattern) Match(id ID) bool {
...
+	for j := nq - 1; j >= 0; j-- {
+		prev[j] = q[j].kind == tokenGap && prev[j+1]
+	}
+	for i := np - 1; i >= 0; i-- {
+		cur[nq] = p[i].kind == tokenGap && prev[nq]
+		for j := nq - 1; j >= 0; j-- {
+			switch {
+			case p[i].kind == tokenGap || q[j].kind == tokenGap:
+				// The gap either ends, or absorbs the other side's next
+				// component.
```
Benehiko: same here

x/secrets/pattern_test.go
```diff
@@ -90,17 +90,63 @@ func TestPatternIncludes(t *testing.T) {
...
+		{"**", "**", true},
+		{"*", "*/foo", false},
+		{"*/foo", "*", false},
+		{"foo/bar", "foo/baz", false},
+		{"foo/*", "foo/bar", true},
+		{"foo/**", "**/foo", true},  // both match "foo"
+		{"bar/**", "foo/**", false}, // first components conflict
+		{"foo/*/baz", "foo/bar/**", true},
+		// Overlap without containment in either direction.
+		{"docker/*/mcp/*", "docker/proj1/**", true}, // both match docker/proj1/mcp/x
```
Benehiko: comment here is redundant

x/secrets/pattern_test.go
```diff
@@ -90,17 +90,63 @@ func TestPatternIncludes(t *testing.T) {
...
+		{"foo/foo/foo/**", "foo/foo/**/foo", true},  // both match foo/foo/foo
```
Benehiko: same here - the test shows they both match

x/secrets/pattern_test.go
```diff
@@ -90,17 +90,63 @@ func TestPatternIncludes(t *testing.T) {
...
+		{"*/foo/**", "**/bar/*", true},              // both match x/foo/bar/y
```
Benehiko: remove comment

x/secrets/pattern_test.go
```diff
@@ -90,17 +90,63 @@ func TestPatternIncludes(t *testing.T) {
...
+		{"docker/mcp/auth/**", "foo/bar", false},
+		// ⟦a/a/**⟧ == ⟦a/a/**/**⟧: equivalent spellings must behave identically.
```
Benehiko: comment either should to explain why or just be removed

x/secrets/pattern_test.go
```diff
@@ -127,14 +173,14 @@ func Test_Filter(t *testing.T) {
 			result: "docker/mcp/auth/foo/bar/*",
 		},
 		{
+			// Overlap without containment (docker/mcp/auth/foo/bar/x) is
+			// not expressible as one of the two patterns.
```
Benehiko: remove

x/secrets/pattern.go
```diff
@@ -77,15 +71,33 @@ func isValidPatternRune(c rune) bool {
...
-// Valid patterns must follow the same validation rules as secret identifiers, with the exception
-// that '*' can be used to match a single component, and '**' can be used to match zero or more components.
+// Pattern matches secret IDs. It follows the ID validation rules, except
+// that '*' matches one component and '**' matches zero or more.
+// Below, matches(p) denotes the set of IDs a pattern p matches.
 type Pattern interface {
-	// Match the [Pattern] against an [ID]
+	// Match reports whether the pattern matches id: O(n*m) time for up to
+	// one '**', O(n*m^k) worst case for k of them, for component counts n
+	// and m.
```
Benehiko:
```suggestion
	// one '**', O(n*m^k) worst case.
```

x/secrets/pattern.go
```diff
@@ -77,15 +71,33 @@ func isValidPatternRune(c rune) bool {
...
-// Pattern can be used to match secret identifiers.
-// Valid patterns must follow the same validation rules as secret identifiers, with the exception
-// that '*' can be used to match a single component, and '**' can be used to match zero or more components.
+// Pattern matches secret IDs. It follows the ID validation rules, except
+// that '*' matches one component and '**' matches zero or more.
+// Below, matches(p) denotes the set of IDs a pattern p matches.
 type Pattern interface {
-	// Match the [Pattern] against an [ID]
+	// Match reports whether the pattern matches id: O(n*m) time for up to
+	// one '**', O(n*m^k) worst case for k of them, for component counts n
```
Benehiko: `...for up to one '**',` doesn't make much sense to me

x/secrets/pattern.go
```diff
@@ -77,15 +71,33 @@ func isValidPatternRune(c rune) bool {
...
-	// Includes returns true if all matches of Pattern [other] are also matches of the current pattern.
-	Includes(other Pattern) bool
-	// String formats the [Pattern] as a string
+	// Contains reports whether the pattern matches every ID [other]
+	// matches:
+	//
+	//	p.Contains(other)  ⇔  matches(other) ⊆ matches(p)
+	//
+	// For example, docker/** contains docker/*/mcp/*, but docker/proj1/**
+	// does not: it misses docker/proj2/mcp/x. Containment implies overlap.
```
Benehiko: `... but docker/proj1/** does not...` confuses me. Does it mean
`docker/** does not contain docker/proj/**` or that `docker/proj/** does not contain docker/*/mcp/*`?

x/secrets/pattern.go
```diff
@@ -77,15 +71,33 @@ func isValidPatternRune(c rune) bool {
...
+	// O(n*m) time, O(m) space, for component counts n and m.
```
Benehiko:
```suggestion
	// O(n*m) time, O(m) space
```

x/secrets/pattern.go
```diff
@@ -77,15 +71,33 @@ func isValidPatternRune(c rune) bool {
...
+	Contains(other Pattern) bool
+	// Overlaps reports whether the pattern and [other] match at least one
+	// ID in common:
+	//
+	//	p.Overlaps(other)  ⇔  matches(p) ∩ matches(other) ≠ ∅
+	//
+	// Symmetric; implied by containment, not the converse: docker/*/mcp/*
+	// and docker/proj1/** overlap yet neither contains the other. O(n*m)
+	// time, O(m) space, for component counts n and m.
+	Overlaps(other Pattern) bool
```
Benehiko, line 102: I think Overlaps needs a bit more explaining or better godocs. Difficult to understand the concept of 

_`docker/*/mcp/*` and `docker/proj1/**` overlap yet neither contains the other._

x/secrets/pattern.go
```diff
@@ -101,35 +113,148 @@ func (p pattern) Match(id ID) bool {
...
+// of IDs exactly when their canonical tokens are equal. Two rewrites give
+// that property:
+//
+//   - In a run of consecutive wildcard components, the '*'s move to the
+//     front and the '**'s merge into one at the end: order within a run
+//     does not affect what it matches, and one '**' already matches any
+//     number of components.
+//   - A pattern consisting only of '**'s becomes "*/**": an ID always has
+//     at least one component, so both match the same IDs, and the first
+//     rewrite already turns the equivalent spelling "**/*" into "*/**".
```
Benehiko:
```suggestion
//   - In a run of consecutive wildcard components, the '*' move to the
//     front and the '**' merge into one at the end: order within a run
//     does not affect what it matches, and one '**' already matches any
//     number of components.
//   - A pattern consisting only of '**' becomes "*/**": an ID always has
//     at least one component, so both match the same IDs, and the first
//     rewrite already turns the equivalent spelling "**/*" into "*/**".
```

x/secrets/pattern.go
```diff
@@ -101,35 +113,148 @@ func (p pattern) Match(id ID) bool {
...
+	}
+	return toks
+}
+
+// covers reports whether p matches every ID q matches, for canonical tokens.
+// The component alphabet is unbounded, so a literal in p never covers an '*'
+// or '**' in q.
+func covers(p, q []token) bool {
+	np, nq := len(p), len(q)
+	// dp[j] == "p[i:] includes q[j:]"; row i needs only row i+1 (prev).
```
Benehiko: don't know what this comment is about

x/secrets/pattern.go
```diff
@@ -101,35 +113,148 @@ func (p pattern) Match(id ID) bool {
...
+	cur := make([]bool, nq+1)
+	prev[nq] = true
+	for i := np - 1; i >= 0; i-- {
+		cur[nq] = p[i].kind == tokenGap && prev[nq]
+		for j := nq - 1; j >= 0; j-- {
+			switch {
+			case p[i].kind == tokenGap:
+				cur[j] = prev[j] || cur[j+1]
+			case q[j].kind == tokenGap:
+				// q's gap may be empty or start with an unknown component.
```
Benehiko: remove

Benehiko (commented):
mostly just godocs 

I think we need to make sure it's super clear to a user using this what it means to use `p.Overlaps` or `p.Contains`

x/secrets/pattern.go
joe0BAB: the later, I'll make it more clear

Benehiko (approved):

PR #645, opened 2026-09-10
fix: validate ExpandID output

joe0BAB:
ExpandID minted its result ID unvalidated, so a pattern carrying a `*` component beside its `**` (e.g. `a/*/**`) produced an ID like `a/*/x` that ParseID rejects and that matches patterns inconsistently. Return the result through ParseID so expansion fails with ErrInvalidID instead of minting an invalid ID.

No current caller can hit this: every realm reaching ExpandID is star-free and parsed IDs cannot contain `*`. ExpandPattern is untouched; mixing `*` there is legitimate and its output is valid by construction.


docker-agent (commented):
### Assessment: 🟢 APPROVE

Benehiko (approved):

PR #646, opened 2026-09-10
docs(pass): use printf in stdin example to avoid trailing newline

joe0BAB:
echo appends a newline that docker pass set stores verbatim, so the documented idiom stored a corrupted secret. Pipe the exact bytes with printf '%s' instead, and drop the pwd.txt intermediate file.


Benehiko (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

docker-agent (commented):
### Assessment: 🟢 APPROVE

Issue #547, opened 2026-06-02
Intermittent Linux keychain failure: DH shared secret stripped of leading zeros ("secret was transferred or encrypted in an invalid way")


Benehiko:
## Summary

The Linux keychain backend (Secret Service over D-Bus, using the `dh-ietf1024-sha256-aes128-cbc-pkcs7` session encryption mode) intermittently fails to create items with:

```
save secret: failed to create item: Couldn't create item: The secret was transferred or encrypted in an invalid way
```

This surfaces as flaky CI on the Linux keychain tests (roughly 1 run in 256).

## Affected code

`store/keychain/internal/go-keychain/secretservice/dh_ietf1024_sha256_aes128_cbc_pkcs7.go`, function `keygenHKDFSHA256AES128`:

```go
sharedSecret, err := group.diffieHellman(theirPublic, myPrivate)
if err != nil {
    return nil, err
}
sharedSecretBytes := sharedSecret.Bytes() // <-- strips leading zero bytes
defer clear(sharedSecretBytes)

r := hkdf.New(sha256.New, sharedSecretBytes, nil, nil)
```

## Root cause

The Secret Service DH key exchange uses the 1024-bit RFC 2409 Second Oakley Group. After computing the Diffie-Hellman shared secret `s = peerPublic^myPrivate mod p`, both peers must derive the AES key with `HKDF-SHA256` over the shared secret encoded as a **fixed-length, 128-byte big-endian** value (the group's prime size).

`math/big.Int.Bytes()` returns the **minimal** big-endian encoding — it drops leading zero bytes. When the shared secret happens to be numerically small enough to have one or more leading zero bytes, our HKDF input is shorter than 128 bytes, while gnome-keyring / libsecret feeds the full zero-padded 128 bytes. The two sides therefore derive **different AES keys**.

With a mismatched key, the item we encrypt cannot be decrypted by the keyring daemon, which rejects it with `The secret was transferred or encrypted in an invalid way`.

The top byte of the shared secret is zero with probability ≈ 1/256, which matches the observed intermittent (~0.4%) failure rate; each additional leading zero byte adds another ~1/256 factor. This is a well-known class of bug in Secret Service client implementations — the same leading-zero issue has been found and fixed in other language bindings.

## Proposed fix

Encode the shared secret into a fixed-length buffer sized to the group prime, using `big.Int.FillBytes` (which left-pads with zeros):

```go
sharedSecret, err := group.diffieHellman(theirPublic, myPrivate)
[...]

PR #652, opened 2026-09-14
fix(keychain): pad DH client public key to fixed length on Linux

### LINKED ISSUE #649: Linux keychain: DH client public key sent unpadded, kwalletd5 rejects ~1/256 sessions ("Client public key size is invalid")
## Summary

`OpenSession` in the Linux keychain store sends the client's Diffie-Hellman public key as `public.Bytes()`. `math/big.Int.Bytes` returns the minimal big-endian encoding and strips leading zero bytes, so about 1 in 256 sessions send a 127-byte key instead of the 128-byte value the 1024-bit Second Oakley Group expects.

kwalletd5 (KDE Frameworks 5.115, Ubuntu 24.04) rejects the short key and `OpenSession` fails:

```
failed to open secretservice session: Client public key size is invalid
```

gnome-keyring and KeePassXC accept the short key, so the failure only shows against KWallet, and only intermittently.

## Where

`store/keychain/internal/go-keychain/secretservice/secretservice.go`, in `OpenSession`:

```go
sessionAlgorithmInput = dbus.MakeVariant(public.Bytes()) // math/big.Int.Bytes is big endian
```

## Related

This is the same class of bug as #547, fixed in 11a73bf ("pad DH shared secret to fixed length on Linux"). That fix padded the *shared secret* fed into HKDF with `FillBytes` into a 128-byte buffer. The *client public key* sent on the wire in `OpenSession` still uses the unpadded `Bytes()` encoding.

## How it was found

Live test in an Ubuntu 24.04 GNOME (Wayland) VM with kwalletd5 owning `org.freedesktop.secrets`. Ten secrets were saved and then fetched one `Get` each. One of about sixty sessions opened during the run failed with the error above. Every other backend and every other session succeeded.

## Fix

Encode the public key into a fixed 128-byte buffer with `FillBytes`, mirroring the shared-secret fix, and add a regression test that pins a key with a leading zero byte to the padded encoding.

```go
publicBytes := make([]byte, 128) // group prime size, matches the shared-secret padding
public.FillBytes(publicBytes)
sessionAlgorithmInput = dbus.MakeVariant(publicBytes)
```

Benehiko:
`OpenSession` sent the DH public key as `public.Bytes()`, which strips leading zero bytes. About 1 in 256 sessions sent 127 bytes instead of the 128 the Second Oakley Group expects, and kwalletd5 rejected the session with "Client public key size is invalid". Same class of bug as #547, which padded only the shared secret.

Two fuzz targets guard the fix. `FuzzEncode` checks that any group value encodes to exactly 128 bytes and round-trips. `FuzzDHExchange` runs a full key exchange over the wire encoding against a peer that enforces the 128-byte length, as kwalletd5 does, and checks both sides derive the same AES key. Seeds include a full-size key whose public value has a leading zero byte. Reverting `encode` to `Bytes()` fails 8 seeds; 60 seconds of fuzzing per target found nothing against the fix.

Fixes #649


docker-agent (commented):
⚠️ **Review did not complete** — The agent ran but did not post a review. [View logs](https://github.com/docker/secrets-engine/actions/runs/34852736266) for details. Re-request a review from `docker-agent` to retry.

joe0BAB (approved):

PR #654, opened 2026-09-15
ci: add fuzz workflow

Benehiko:
Adds a `Fuzz` workflow that runs every Go fuzz target with a time budget. The repo has two targets today (`FuzzEncode`, `FuzzDHExchange` in `secretservice`) and nothing ran them beyond their seed corpus.

| Change | Effect |
| --- | --- |
| `make fuzz` (`FUZZTIME` default 30s) | Finds `Fuzz*` functions outside `vendor/`, confirms them with `go test -list`, runs each on its own. New targets need no workflow edit. |
| Triggers | PR and push to `main`: 30s per target. Nightly: 10m. `workflow_dispatch`: `fuzztime` input. |
| Corpus cache | Restores the newest `fuzz-corpus-*` cache and saves after every run, so runs build on earlier findings. |
| Failure artifact | Uploads `testdata/fuzz/**` as `fuzz-failing-inputs` so a crash can be reproduced locally and committed as a regression seed. |
| `docs/ci/fuzz.md` | Triggers, corpus cache, and reproduction steps. |

Verified locally: `make fuzz FUZZTIME=3s` runs both targets, a bad duration exits non-zero, actionlint reports no findings.


docker-agent (commented):
### Assessment: 🟢 APPROVE

joe0BAB (approved):

PR #657, opened 2026-09-23
feat(resolver): omit the Authorize expiry when the decision never expires

joe0BAB:
A zero Expiry on secrets.AuthorizeResponse now means the decision never expires. The handler leaves expires_at absent for a zero Expiry, and the client decodes an absent expires_at as the zero time instead of the Unix epoch, which read as already expired. Document the meaning on the field and on client.Authorize, and cover the finite and the absent expiry on both sides.


Benehiko (approved):

docker-agent (commented):
### Assessment: 🟢 APPROVE

PR #659, opened 2026-09-23
refactor(api): clarify Desktop and standalone socket paths

Benehiko:
`DefaultSocketPath()` names Docker Desktop's endpoint, while `client.New()` uses `DaemonSocketPath()` for the standalone engine. The names make the SDK default and the two installations difficult to distinguish.

Add `DesktopSocketPath()` and `StandaloneSocketPath()`, and keep the old functions as deprecated forwarding aliases. Update SDK, plugin, and pass callers to use the explicit names. Document the platform paths, current endpoint selection, and how to connect the SDK to Docker Desktop. Existing socket addresses and default selection are preserved; automatic discovery remains a separate change.

Validation:

- `go test -race ./x/api ./client/... ./plugins/pass/commands`
- Linux and Windows cross compilation of `./x/api ./client/... ./plugin/... ./plugins/pass/commands`
- macOS plugin build and `git diff --check`


docker-agent (commented):
### Assessment: 🟢 APPROVE

plugins/pass/commands/run.go
```diff
@@ -78,7 +78,7 @@ func WithResponseTimeout(responseTimeout time.Duration) RunOption {
 	}
 }
 
-// WithSocketPath overrides the engine socket path; empty means the default.
+// WithSocketPath overrides the engine socket path; empty uses [api.DesktopSocketPath].
```
Benehiko, line 81: @joe0BAB looks like RunOptions will need to return an error. Will do that in a follow-up PR since it's a breaking change

joe0BAB (approved):

PR #661, opened 2026-09-23
feat(pass): authorize every se:// reference before resolving

joe0BAB:
`docker pass run` now parses the merged environment once into typed entries and asks the engine for one Authorize decision covering every `se://` reference before it resolves any of them. The access-control prompt therefore appears once for the whole set instead of once per reference, and a deny stops the command before the child starts.

Cover parsing, resolution, and the authorize-then-resolve order with unit tests and two runs against a mock engine on a unix socket.


docker-agent (commented):
### Assessment: 🟢 APPROVE

The authorize-before-resolve pattern is correctly implemented: the merged environment is parsed once into typed `envVar` entries, `Authorize` is called with all `se://` patterns in a single batch before any secret is resolved, and a deny response halts the command before the child process starts. The core logic and tests are sound.

#### Lower-confidence findings (not posted inline)

These were identified by static analysis but are low-severity and unverified. They are surfaced for author awareness only and do not affect the approval.

- **[low] `plugins/pass/commands/run.go`:279 — `resolveEnv` returns nil slice when `vars` is empty, causing child to inherit parent environment** (unverified, low)
  In Go's `exec` package, `cmd.Env = nil` means the child inherits the parent's full environment, while `cmd.Env = []string{}` starts the child with an empty environment. The new `resolveEnv` uses `var out []string` (nil); the previous implementation used `make([]string, 0, len(env))` (non-nil empty). If `vars` is ever empty (e.g., a fully cleared `--env-file`), the behavior silently diverges. In practice `os.Environ()` always populates at least a few entries, making this a hypothetical edge case. Fix if strict semantics are required: `out := make([]string, 0, len(vars))`.

- **[low] `plugins/pass/commands/run.go`:262 — `authorizeEnv` sends duplicate patterns when the same `se://` reference appears in multiple env vars** (unverified, low)
  If two env vars both resolve to the same pattern (e.g., `TOKEN=se://gh-token` and `BACKUP_TOKEN=se://gh-token`), the `Authorize` call receives `[gh-token, gh-token]`. Whether the engine treats this as one resource or two depends on engine policy. If de-duplication is desired, collect patterns into a `map[string]secrets.Pattern` before calling `Authorize`.

- **[low] `plugins/pass/commands/run_test.go`:344 — no end-to-end test for the all-plain-env fast path through `RunCommand`** (unverified, low)
  The two new integration tests both use an `se://` reference. A regression in the `len(patterns) == 0` early-return path of `authorizeEnv` (e.g., an accidental `Authorize` call on an empty pattern list) would not be caught by the integration tests. The unit test for `authorizeEnv` covers this path, but end-to-end coverage through `RunCommand` is absent.

docker-agent (commented):
### Assessment: 🟢 APPROVE

Benehiko (approved):

